PR #4, opened 2025-01-13
fix: speed up stream objects creation

nqkdev:
Hi, thank you for the great plugin. 

I noticed it took some time to populate the streaming links when used with mediaflow-proxy. Therefore I propose a small change to make the experience a bit more seamless.


Viren070 (approved):

Viren070:
Thank you!

PR #202, opened 2025-06-20
fix: overlapping snippets modal

davidemarcoli:
On Firefox the snippets modal was overlapping

My solution changes the behavior to break-all and not only words. Breaking on words is nicer to look at but unfortunately not responsive when using the formatter syntax that doesn't contain many spaces.

Before:
image

After:
image


Viren070 (approved):

Viren070:
Thanks!

PR #225, opened 2025-06-26
Update .env.sample with correct database uri for docker

ridvan-cln:
## Summary by CodeRabbit

* **Chores**
  * Updated the sample environment configuration to correct the SQLite database path and added a newline at the end of the file.


Viren070:
The default also works, as the `.` is resolved to the current directory (`/app`). There isn't a need to change it to `sqlite:///app/data...`

PR #273, opened 2025-07-21
improved german language detection + prevent false positives + improved overall result

mindgam3s:
in Portuguese and Spanish the 'de' is used as a preposition like 'from/of'  ...

 (which is in a lot of movie/episode titles)

 ... so the filter would find those as well 😦
 
 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of German language tags for greater accuracy when parsing text, ensuring more reliable identification across different formats.


mindgam3s:
@Viren070 
couldn't deploy it to test it myself, but if I understand your code correctly that should work, please leave feedback if necessary

Issue #355, opened 2025-09-06
bug: conditional doesn’t work anymore


oAmarilha:
### Deployment Method

Private ElfHosted Instance

### Addon Version

2025.09.06.1314-nightly

### Bug Area

Configuring

### Bug Description / Steps to Reproduce

After the changes from https://github.com/Viren070/AIOStreams/pull/335 the ~ 'contains' is not working anymore in formatter. Example: {streams.languages::~Portuguese["🇧🇷 "||""]}
The example is returning false, even if there is Portuguese in the languages array.

### Expected Behavior

If a string or list contains the string typed, it should return true, instead of false.

### Configuration Export

[aiostreams-config.json](https://github.com/user-attachments/files/22188405/aiostreams-config.json)

### Debugging Checklist

- [ ] If applicable, I have included server logs 
- [ ] If applicable, I have included MediaFlow logs

### Issue Checklist

- [x] I have written a short but informative title that clearly describes the issue.
- [x] I have given clear and descriptive steps to reproduce the issue.
- [x] I have checked open and closed issues and confirmed that this is not a duplicate of another issue.
- [x] I have filled out all of the requested information adequately. 
- [x] I am using the [latest version](https://github.com/Viren070/AIOStreams/releases/latest).

oAmarilha:
![image](https://github.com/user-attachments/assets/3dfe841c-b887-451c-8a82-ceb8ff49a530)
Example from pull request is returning false even with English in the languages array.

oAmarilha:
![image](https://github.com/user-attachments/assets/aa10710e-b802-402c-a0c9-181e78f5d888)
Conditional equal is also returning false even if the string matches. Example from pull request was used.

PR #361, opened 2025-09-06
[Bugfix-#355] Update Conditional check to properly typecheck

DavidGracias:
Conditionals have been failing due to
- Improperly configured .toString / .toLowerCase / etc (forcing only human input `check` to be lower but not `value`)
- Improperly configured comparison operators for numeric types (causing issues like "2" < "12" to be false when we'd actually want it to be 2 < 12, true)

Updating with proper types and updating logic a bit to cover corner cases that caused this to fail.
Closes https://github.com/Viren070/AIOStreams/issues/355

image

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate conditional evaluation with explicit numeric comparisons when values are numeric.
  * Resolved inconsistent whitespace checks; presence detection now uses non-whitespace tests.
  * Conditional branches now receive richer parsing context, improving resolution of true/false branches.

* **Refactor**
  * Comparisons are now case-sensitive by default, improving strictness.
  * Prefix-selection and pattern-generation order changed — may alter matching in some edge cases.

### DIFF
--- a/packages/core/src/formatters/base.ts
+++ b/packages/core/src/formatters/base.ts
@@ -551,17 +551,20 @@ export abstract class BaseFormatter {
         
         // PREFIX
         else if (isPrefix) {
+          // get the longest prefix match
           const modPrefix = Object.keys(conditionalModifiers.prefix).sort((a, b) => b.length - a.length).find(key => mod.startsWith(key))!!;
           
+          // Pre-process string value and check to allow for intuitive comparisons
           const stringValue = value.toString().toLowerCase();
           let stringCheck = mod.substring(modPrefix.length).toLowerCase();
-          // remove spaces/tabs/newlines from stringCheck if they aren't in stringValue
+          // remove whitespace from stringCheck if it isn't in stringValue
           stringCheck = !/\s/.test(stringValue) ? stringCheck.replace(/\s/g, '') : stringCheck;
           
-          const [numericValue, numericCheck] = [Number(value), Number(stringCheck)]
-          // use numeric comparison if possible, otherwise use string (value, check) params
-          const [paramValue, paramCheck] = ["<", "<=", ">", ">="].includes(modPrefix) && 
-            !isNaN(numericValue) && !isNaN(numericCheck) ? [numericValue, numericCheck] : [stringValue, stringCheck]
+          const isNumericComparison = ["<", "<=", ">", ">="].includes(modPrefix) && 
+            !Number.isNaN(Number(stringValue)) && !Number.isNaN(Number(stringCheck));
+          const [paramValue, paramCheck] = isNumericComparison 
+            ? [Number(stringValue), Number(stringCheck)] 
+            : [stringValue, stringCheck];
           
           conditional = conditionalModifiers.prefix[modPrefix as keyof typeof conditionalModifiers.prefix](
             paramValue as any, 
--- a/packages/core/src/formatters/base.ts
+++ b/packages/core/src/formatters/base.ts
@@ -223,10 +223,10 @@ export const conditionalModifiers = {
     '^': (value: string, check: string) => value.endsWith(check),
     '~': (value: string, check: string) => value.includes(check),
     '=': (value: string, check: string) => value == check,
-    '>=': (value: string, check: string) => value >= check,
-    '>': (value: string, check: string) => value > check,
-    '<=': (value: string, check: string) => value <= check,
-    '<': (value: string, check: string) => value < check,
+    '>=': (value: string | number, check: string | number) => value >= check,
+    '>': (value: string | number, check: string | number) => value > check,
+    '<=': (value: string | number, check: string | number) => value <= check,
+    '<': (value: string | number, check: string | number) => value < check,
   },
 }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/formatters/base.ts
```diff
@@ -211,22 +211,22 @@ export const conditionalModifiers = {
...
-      if (typeof value === 'string') return (value.replace(/ /g, '').length > 0);
+      if (typeof value === 'string') return value.replace(/ /g, '').length > 0;
       if (Array.isArray(value)) return value.length > 0;
       // For other types (numbers, booleans, objects), consider them as "existing"
       return true;
     },
   },
 
   prefix: {
-    '$': (value: any, check: any) => value.toLowerCase().startsWith(check.toLowerCase()),
```
DavidGracias, line 222: Change to strings to enforce type elsewhere

packages/core/src/formatters/base.ts
```diff
@@ -553,16 +553,14 @@ export abstract class BaseFormatter {
...
-          for (let key of Object.keys(conditionalModifiers.prefix)) {
-            if (mod.startsWith(key)) {
-              var checkKey = mod.substring(key.length);
-              if (typeof value !== 'string' || !value.includes(' ')) {
-                checkKey = checkKey.replace(/ /g, '');
-              }
-              conditional = conditionalModifiers.prefix[key as keyof typeof conditionalModifiers.prefix](value, checkKey);
-              break;
-            }
+          const modPrefix = Object.keys(conditionalModifiers.prefix).find(key => mod.startsWith(key))!!;
```
DavidGracias: find: Returns the value of the first element in the array where predicate is true, and undefined otherwise.

@param predicate
find calls predicate once for each element of the array, in ascending order, until it finds one where predicate returns true. If such an element is found, find immediately returns that element value. Otherwise, find returns undefined.

@param thisArg
If provided, it will be used as the this value for each invocation of predicate. If it is not provided, undefined is used instead.

DavidGracias: We know one exists from isPrefix definition

packages/core/src/formatters/base.ts
```diff
@@ -553,16 +552,22 @@ export abstract class BaseFormatter {
...
-            if (mod.startsWith(key)) {
-              var checkKey = mod.substring(key.length);
-              if (typeof value !== 'string' || !value.includes(' ')) {
-                checkKey = checkKey.replace(/ /g, '');
-              }
-              conditional = conditionalModifiers.prefix[key as keyof typeof conditionalModifiers.prefix](value, checkKey);
-              break;
-            }
-          }
+          const modPrefix = Object.keys(conditionalModifiers.prefix).sort((a, b) => b.length - a.length).find(key => mod.startsWith(key))!!;
```
DavidGracias, line 555: Get longest matching prefix

packages/core/src/formatters/base.ts
```diff
@@ -553,16 +552,22 @@ export abstract class BaseFormatter {
...
+          
+          const stringValue = value.toString().toLowerCase();
+          let stringCheck = mod.substring(modPrefix.length).toLowerCase();
+          // remove spaces from stringCheck if spaces aren't in stringValue
+          stringCheck = !stringValue.includes(' ') ? stringCheck.replace(/ /g, '') : stringCheck;
+          
+          const [numericValue, numericCheck] = [Number(value), Number(stringCheck)]
+          // use numeric comparison if possible, otherwise use string (value, check) params
```
DavidGracias, line 563: Allows for proper comparisons

Example: casting "12" to 12 and "2" to 2 so that "12" < "2" (true) becomes 12 < 2 (false) and evaluates correctly

packages/core/src/formatters/base.ts
DavidGracias: Also adding pre-conditioning to params elsewhere (all strings are lower-cases, remove non-whitespace chars from check if none exist in value, ...)

packages/core/src/formatters/base.ts
```diff
@@ -417,10 +417,8 @@ export abstract class BaseFormatter {
     const variable = `(?<variableName>${validVariables.join('|')})\\.(?<propertyName>${validProperties.join('|')})`;
 
     // Dynamically build the `modifier` regex pattern from modifier keys
-    // Sort by length (longest first) to ensure more specific patterns match before shorter ones
```
DavidGracias, line 420: Not needed here, removing for the sake of simplicity

packages/core/src/formatters/base.ts
```diff
@@ -553,16 +551,22 @@ export abstract class BaseFormatter {
...
-          for (let key of Object.keys(conditionalModifiers.prefix)) {
-            if (mod.startsWith(key)) {
-              var checkKey = mod.substring(key.length);
-              if (typeof value !== 'string' || !value.includes(' ')) {
-                checkKey = checkKey.replace(/ /g, '');
-              }
-              conditional = conditionalModifiers.prefix[key as keyof typeof conditionalModifiers.prefix](value, checkKey);
-              break;
-            }
-          }
```
DavidGracias, line 565: Below:

modPrefix matches the longest matching prefix modifier (i.e. <,>,...), which we know it exists due to isPrefix variable being true


stringValue/stringCheck are pre-processed to be simple strings (lowercased and match on having/not having whitespace)


conditional is updated with paramValue/paramCheck based off if we're able to case these values to a valid number type and we can accept that type in the function or not

PR #368, opened 2025-09-09
Allow Multiple Modifiers

DavidGracias:
Update modifiers to allow for multiple back-to-back modifiers, for example:

{stream.languages::first::title} should now properly work, while still allowing for single modifiers


packages/core/src/formatters/base.ts
```diff
@@ -431,7 +440,7 @@ export abstract class BaseFormatter {
     // TZ Locale pattern (e.g. 'UTC', 'GMT', 'EST', 'PST', 'en-US', 'en-GB', 'Europe/London', 'America/New_York')
     const modTZLocale = `::(?<mod_tzlocale>[A-Za-z]{2,3}(?:-[A-Z]{2})?|[A-Za-z]+?/[A-Za-z_]+?)`;
 
-    const regexPattern = `\\{${variable}(${modifier})?(${modTZLocale})?(${checkTF})?\\}`;
+    const regexPattern = `\\{${variable}(?<modifiers>(${singleValidModifier})+)?(${modTZLocale})?(${checkTF})?\\}`;
```
DavidGracias, line 487: (${singleValidModifier})+ captures any singleValidModifier 1 or more times

(?<modifiers>...)? makes the group `modifiers` optional

packages/core/src/formatters/base.ts
```diff
@@ -585,39 +601,35 @@ export abstract class BaseFormatter {
     }
 
     // --- STRING MODIFIERS ---
-    if (typeof value === 'string') {
-      if (mod in stringModifiers) return stringModifiers[mod as keyof typeof stringModifiers](value);
-      return `{unknown_str_modifier(${mod})}`;
```
DavidGracias, line 590: nit: make returns consistently on next line & return `{unknown_<var type>_modifier(${mod})}`; returns at the end if we're unable to find the modifier for that variable.property type

Helpful debugging for users in cases like:
{stream.filename::random} -> will print `{unknown_string_modifier(random)}` instead of `{unknown_modifier(random)}`

packages/core/src/formatters/base.ts
```diff
@@ -585,39 +601,35 @@ export abstract class BaseFormatter {
...
 
     // --- ARRAY MODIFIERS ---
     else if (Array.isArray(value)) {
-      if (mod in arrayModifiers) {
-        const result = arrayModifiers[mod as keyof typeof arrayModifiers](value);
-        if (typeof result === 'string') return result;
-        return result.join(', ');
-      }
+      if (mod in arrayModifiers)
+        return arrayModifiers[mod as keyof typeof arrayModifiers](value).toString();
```
DavidGracias: using .toString() instead of casting to string myself

packages/core/src/formatters/base.ts
```diff
@@ -104,10 +104,11 @@ export interface ParseValue {
   debug?: {
     json: string | null;
     jsonf: string | null;
+    modifier: string | null;
```
DavidGracias, line 107: Add new debugging string `{debug.modifer}` to show basic implementations of each modifier and how they compare

packages/core/src/formatters/base.ts
```diff
@@ -244,14 +245,48 @@ const hardcodedModifiersForRegexMatching = {
...
+  ...hardcodedModifiersForRegexMatching,
   ...stringModifiers,
   ...numberModifiers,
   ...arrayModifiers,
   ...conditionalModifiers.exact,
   ...conditionalModifiers.prefix,
-  ...hardcodedModifiersForRegexMatching,
 }
 
+const debugModifierToolReplacement = `
```
DavidGracias, line 257: String replacement for `{debug.modifier}` -- good for users to see as an example and for us to debug in the future

packages/core/src/formatters/base.ts
```diff
@@ -402,6 +437,13 @@ export abstract class BaseFormatter {
     };
   }
 
+  // Build the modifier regex pattern from modifier keys
+  protected buildModifierRegexPattern(): string {
+    const validModifiers = Object.keys(modifiers)
+      .map(key => key.replace(/[\(\)\'\"\$\^\~\=\>\<]/g, '\\$&'));
+    return `::(${validModifiers.join('|')})`;
+  }
```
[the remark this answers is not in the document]
DavidGracias: Those aren't needed for now. The less we overly replace, the better

Viren070 (approved):

PR #369, opened 2025-09-10
[Bug Fix] Update Multiple Modifiers Feature to properly parse modifiers, check_tf, etc...

DavidGracias:
Prior to this, we weren't properly capturing each individual modifier and instead were incorrectly clumping in the check_tf into the modifier and/or not properly terminating the regex string, making modifiers break!

This fix should resolve these issues. I would love reviewers to help ensure this fix looks good!

## Summary by CodeRabbit

- New Features
  - Clearer output when using unknown modifiers (now rendered explicitly as unknown).
- Bug Fixes
  - More robust handling of missing add-on details to prevent runtime errors.
- Refactor
  - Reworked modifier processing into a modular flow for improved reliability.
  - Dynamic detection of available template variables replaces static lists.
- Chores
  - Narrowed support for certain prefix-style modifiers, which are no longer recognised.

### DIFF
--- a/packages/core/src/formatters/base.ts
+++ b/packages/core/src/formatters/base.ts
@@ -104,6 +104,7 @@ export interface ParseValue {
   debug?: {
     json: string | null;
     jsonf: string | null;
+    modifier: string | null;
   };
 }
 
@@ -164,6 +165,7 @@ const hardcodedParseValueKeysForRegexMatching: ParseValue = {
   debug: {
     json: null,
     jsonf: null,
+    modifier: null,
   },
 }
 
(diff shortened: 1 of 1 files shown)


packages/core/src/formatters/base.ts
```diff
@@ -97,9 +97,9 @@ export interface ParseValue {
     cached: boolean | null;
   };
   addon?: {
-    name: string;
```
DavidGracias, line 100: nit: this nullability check allows for hardcodedParseValueKeysForRegexMatching to be deleted and for us to use 
`const hardcodedParseValueKeysForRegexMatching = this.convertStreamToParseValue({} as ParsedStream);` instead

packages/core/src/formatters/base.ts
```diff
@@ -536,24 +476,16 @@ export abstract class BaseFormatter {
         continue;
       }
 
-      // Process - Modifier(s)
+      // Validate and Process - Modifier(s)
```
DavidGracias, line 479: Refactoring:

1. Move `{unknown_...` logic OUT of modifier function (handle errors at this level instead)
2. Move conditional logic into multiple modifier (applyModifiers) level to allow for handling AFTER all changes have been processed
3. nit: changed value/property variablename to `input` to be more descriptive

packages/core/src/formatters/base.ts
```diff
@@ -578,37 +510,69 @@ export abstract class BaseFormatter {
       .replace(/\{tools.newLine\}/g, '\n');
   }
 
-  protected modifier(
+  protected applyModifiers(
+    groups: {[key: string]: string},
+    input: any,
+    parseValue: ParseValue,
+  ): string | undefined {
+    const singleModTerminator = '((::)|($))'; // :: if there's multiple modifiers or $ for the end of the string
```
DavidGracias, line 518: This terminator is more or less the fix. We need to capture the modifier group by ensuring we are capturing the full modifier (previously, ::=X only captured ::=)

packages/core/src/formatters/base.ts
```diff
@@ -578,37 +510,69 @@ export abstract class BaseFormatter {
...
+        modMatch[1], // First capture group (the modifier name)
+        result,
+        groups.mod_tzlocale ?? "",
+      );
+    }
+
+    // handle unknown modifier result
+    switch (typeof result) {
+      case 'undefined': return undefined;
+      case 'boolean':
```
DavidGracias, line 535: Once all modifiers have been applied, handle conditional at top level

DavidGracias:
@coderabbitai review

PR #379, opened 2025-09-14
feat(http): re-use proxy agents

webstreamr:
I can't back it up, except with semi-trustable AI output, but re-using proxy agents, instead of re-creating them with every request, could improve performance in some situations. I'm especially thinking about e.g. torrentio on popular public instances.


webstreamr:
sorry for the exceeding the rate limit here. my local clone was not fully configured properly with git author info and signing, so I force-pushed until I got it right 🙈

PR #384, opened 2025-09-18
[Formatter Feature] 1) Allow {...} to contain whitespace for readability 2) Allow nested conditionals within conditionals

DavidGracias:
1) Allows whitespace in template definition between curly braces
- Matching manually rather than via regex for speed and customization ability (including this handy benefit!)
Example (all versions below work)

2) Allows for conditionals to contain full variables (i.e. `{variableType1.propertyName1::modifier1a:modifier1b::comparator1and2::variableType2.propertyName2::modifier2["{variableType3.propertyName3::modifier3["nested true"||"nested false"]"|""]}`

## Summary by CodeRabbit

- New Features
  - Support for nested variables and modifiers in templates.
  - Case-insensitive matching for template keys.
  - New conditional and timezone/locale modifiers.
  - New newline and remove-line tokens for cleaner output.

- Performance
  - Faster template compilation with parallel processing for name/description.
  - More modular compilation for improved scalability.

- Bug Fixes
  - More consistent error handling during variable parsing.
  - Improved handling of duplicate keys and nested resolution.

- Refactor
  - Overhauled parsing and regex logic to enhance reliability and maintainability.

### DIFF
--- a/packages/core/src/formatters/base.ts
+++ b/packages/core/src/formatters/base.ts
@@ -104,30 +104,30 @@ export interface ParseValue {
     json: string | null;
     jsonf: string | null;
   } & typeof DebugToolReplacementConstants;
+  tools: {
+    removeline: string;
+    newline: string;
+  };
 }
 
 /**
  * Pre-compiled function that takes ParseValue and returns formatted string
  */
-type CompiledParseFunction = (parseValue: ParseValue) => string;
-type CompiledVariableWInsertFn = {
-  resultFn: (parseValue: ParseValue) => ResolvedVariable;
-  insertIndex: number;
-};
+type ParseValueToString = (parseValue: ParseValue) => string;
 /**
- * Pre-compiled function that takes ParseValue and returns `ResolvedVariable` (future: and the variable's context for caching purposes)
+ * Pre-compiled function that takes ParseValue and returns `ResolvedVariable`
  *
  * Retrieves the resolved variable (including modifiers) given a ParseValue (e.g. `stream.cached:istrue` -> `{result: true}` or `stream.languages::istrue` -> `{error: "unknown_array_modifier(istrue)"}`)
  */
-type CompiledModifiedVariableFn = (parseValue: ParseValue) => ResolvedVariable;
+type ParseValueToVariable = (parseValue: ParseValue) => ResolvedVariable;
 
 export abstract class BaseFormatter {
   protected config: FormatterConfig;
   protected userData: UserData;
 
   private regexBuilder: BaseFormatterRegexBuilder;
-  private precompiledNameFunction: CompiledParseFunction | null = null;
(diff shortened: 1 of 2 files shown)


packages/core/src/formatters/base.ts
```diff
@@ -104,30 +104,30 @@ export interface ParseValue {
     json: string | null;
     jsonf: string | null;
   } & typeof DebugToolReplacementConstants;
+  tools: {
```
DavidGracias, line 107: Add tools to parseValue so that it's not as "hardcoded" and rather a more built in variableTemplate

packages/core/src/formatters/base.ts
```diff
@@ -341,127 +325,328 @@ export abstract class BaseFormatter {
       );
     }
 
-    const placeHolder = " ";
-
-    // Iterate through all {...} matches
-    while ((matches = re.exec(str))) {
```
DavidGracias: `while ((matches = re.exec(str))) {` replaced with manual variable checking (i.e. going through string character by character) and moved below logic to `compileTemplateHelper`

packages/core/src/formatters/base.ts
```diff
@@ -341,127 +325,328 @@ export abstract class BaseFormatter {
...
-          resultStr.slice(insertIndex+placeHolder.length);
+          resultStr.slice(insertedIndex + PLACEHOLDER.length);
       }
-
-      return resultStr
+      return resultStr;
     };
   }
 
+  /**
```
DavidGracias, line 572: The existing logic that we know and love, moved into this helper functions (this was the core parsing logic in the previous while loop)

At this point, we can assume modifiedVariable is valid, has no unexpected whitespace, and has all nested variables resolved!

hugo9655:
would nested conditionals make something like this possible?
`{stream.encode::exists["📹 {stream.encode}{stream.visualTags::exists[" • {stream.visualTags::join(' • ')}"||""]}"||"{stream.visualTags::exists["📹 {stream.visualTags::join(' • ')}"||""]}"]}`

Viren070:
> would nested conditionals make something like this possible? `{stream.encode::exists["📹 {stream.encode}{stream.visualTags::exists[" • {stream.visualTags::join(' • ')}"||""]}"||"{stream.visualTags::exists["📹 {stream.visualTags::join(' • ')}"||""]}"]}`

@hugo9655, yes, but this is currently possible with conditionals, which was added in #381 

```
{stream.encode::exists["📹 {stream.encode}"||""]}{stream.visualTags::exists::and::stream.encode::exists["• {stream.visualTags::join(' • ')}"||""]}{stream.encode::exists::isfalse::and::stream.visualTags::exists["📹 {stream.visualTags::join(' • ')}"||""]}
```

adding nested conditionals does make it more intuitive however and allow for more options.

hugo9655:
ohh i see, cool then.

PR #386, opened 2025-09-18
Format base.ts AND add replace("needle", "replaceValue") modifier for strings

DavidGracias:


packages/core/src/formatters/base.ts
```diff
@@ -515,19 +616,39 @@ export abstract class BaseFormatter {
     // --- STRING MODIFIERS ---
     else if (typeof variable === 'string') {
       if (mod in ModifierConstants.stringModifiers)
-        return ModifierConstants.stringModifiers[mod as keyof typeof ModifierConstants.stringModifiers](variable);
+        return ModifierConstants.stringModifiers[
+          mod as keyof typeof ModifierConstants.stringModifiers
+        ](variable);
+
+      // handle hardcoded modifiers here
```
DavidGracias, line 624: Functional change -- adding replace as a hardcoded modifier here

packages/core/src/formatters/base.ts
```diff
@@ -697,20 +834,24 @@ class ModifierConstants {
       '<=': (value: string | number, check: string | number) => value <= check,
       '<': (value: string | number, check: string | number) => value < check,
     },
-  }
+  };
 
-  static hardcodedModifiersForRegexMatching = {
+	static hardcodedModifiersForRegexMatching = {
```
DavidGracias, line 840: Functional change: adding all "versions" of `replace` that I can think of in regex

packages/core/src/formatters/base.ts
```diff
@@ -306,84 +318,131 @@ export abstract class BaseFormatter {
...
+          };
         };
       } // end of CHECK TRUE/FALSE logic
 
-      str = str.slice(0, index) + str.slice(re.lastIndex);
-      re.lastIndex = index;
-      compiledMatchTemplateFns.push({ resultFn: precompiledResolvedVariableFn, insertIndex: index });
-    } // end of while loop
-
+      str = str.slice(0, index) +placeHolder+ str.slice(re.lastIndex);
```
DavidGracias, line 423: Adding placeholder (could be any string with length>0, I chose " ") to prevent variable ordering from being messed up!

Viren070 (approved):

PR #389, opened 2025-09-19
[Bugfix] Fix Formatter `{tools.removeLine}` and other post-processing of string

DavidGracias:
# Properly post-processed string (removing line properly when needed)
Screenshot 2025-09-19 at 6 05 52 PM

---
# Properly post-process string (not removing line if not needed)
Screenshot 2025-09-19 at 6 07 13 PM

### DIFF
--- a/packages/core/src/formatters/base.ts
+++ b/packages/core/src/formatters/base.ts
@@ -311,7 +311,22 @@ export abstract class BaseFormatter {
   }
 
   protected async compileTemplate(str: string): Promise<CompiledParseFunction> {
-    if (!str) return () => '';
+    const compiledHelper = await this.compileTemplateHelper(str);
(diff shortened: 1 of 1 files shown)


packages/core/src/formatters/base.ts
```diff
@@ -311,7 +311,22 @@ export abstract class BaseFormatter {
   }
 
   protected async compileTemplate(str: string): Promise<CompiledParseFunction> {
-    if (!str) return () => '';
+    const compiledHelper = await this.compileTemplateHelper(str);
+    return (parseValue: ParseValue) => {
```
DavidGracias, line 315: Make this happen only once, not on each recursive call (i.e. in the below `compileTemplateHelper` calls)

PR #431, opened 2025-10-13
Feat: Are You Still there Functionality

### LINKED ISSUE #425: feature request: "Are you still there?" / "Still watching?" workaround
### Checklist

Hello,

First of all, thank you for the incredible work on this amazing add-on, it really revolutionized the way I use Stremio. I’ve been using Stremio for the past five years, and there’s just one feature I feel is missing to make it perfect: the “Are you still there?” function. My wife often falls asleep while watching shows like Friends, HIMYM, or The Good Place, and I’d really appreciate a way to stop the autoplay after some time, so it doesn’t run all night long.

### Proposed Solution

I see that recently you've implemented something around the autoplay, I was just wondering if it would be possible to code some way that after a couple of episodes, the autoplay returns nothing then the player would go back to the stream selection screen and ultimately the device would stop and lock itself. Of course, sometimes the feature would trigger unwillingly but it won't be a problem to select the stream from time to time

For now, I’ve been using the “Focus” feature on Android to set a timed screen lock, but that’s more of a workaround than a solution.

Really don't know if this is feasible or not but wanted to ask you guys, thanks again for the hard work.

nhyyebo:
in which you can set how many episodes will autoplay before it stops, within a specific time period (cooldown).

Cooldown sets how far apart episode starts can be and still count toward the N. If a gap exceeds cooldown, the counter resets and wont trigger.

closes #425

## Summary by CodeRabbit

* **New Features**
  * Added an “Are you still there?” per-user, per‑series autoplay check that pauses autoplay after a configurable number of consecutive episodes within a cooldown window.
  * New Pro settings let you enable the feature and set episodes‑before‑check and cooldown‑minutes; sensible defaults and minimum values are enforced.


packages/server/src/routes/stremio/stream.ts
```diff
@@ -39,14 +40,52 @@ router.get(
...
+      ) {
+        const cfg = req.userData.areYouStillThere;
+        const threshold = cfg.episodesBeforeCheck ?? 3;
+        const cooldownMs = (cfg.cooldownMinutes ?? 60) * 60 * 1000;
+        const cache = Cache.getInstance<
+          string,
+          { count: number; lastAt: number }
+        >('areYouStillThere', 10000);
+        // Use series identifier only (before first ':') so counts are per show
+        const seriesId = id.split(':')[0] || id;
```
Viren070: We could use https://github.com/Viren070/AIOStreams/blob/main/packages/core/src/utils/id-parser.ts instead to parse the ID, Currently this would parse tmdb/tvdb IDs incorrectly e.g. `tmdb:2144:2:1` gets the `seriesId` parsed as `tmdb`.

nhyyebo: oops totally overlooked that. on it.

PR #481, opened 2025-11-10
cleanTitle(): replace umlaut characters with ASCII equivalents

nexariton:
Update the cleanTitle() function in utils.ts to replace umlaut characters with their ASCII transliterations.
This is the most common naming convention on indexers so this is needed to get good results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced title processing to correctly handle German special characters and umlauts (Ä, ä, Ö, ö, Ü, ü) as well as the ß symbol by converting them to ASCII equivalents. This ensures consistent and reliable text normalisation across the platform.

### DIFF
--- a/packages/core/src/parser/utils.ts
+++ b/packages/core/src/parser/utils.ts
@@ -84,7 +84,11 @@ export function normaliseTitle(title: string) {
 }
 
 export function cleanTitle(title: string) {
-  let cleaned = title.normalize('NFD');
+  const umlautMap: Record<string,string> = {
+    'Ä':'Ae','ä':'ae','Ö':'Oe','ö':'oe','Ü':'Ue','ü':'ue','ß':'ss'
+  };
+  // replace German umlauts with ASCII equivalents, then normalize to NFD
+  let cleaned = title.replace(/[ÄäÖöÜüß]/g, c => umlautMap[c]).normalize('NFD');
 
   for (const char of ['♪', '♫', '★', '☆', '♡', '♥', '-']) {
     cleaned = cleaned.replaceAll(char, ' ');


be-na17:
I don't know that this is a good idea. Most (usenet) indexers already handle this case internally. And for those that do require strict matching you would no longer get results for filenames containing the umlauts.

nexariton:
Aiostreams currently replaces öäü with oau so the information is lost before the indexer can handle it.
From my observations this pr improves the results significantly.
Are you suggesting that aio should use the unaltered umlaut containing title for querys to usenet indexers because they can handle it better? That would be an additional discussion.

be-na17:
> Aiostreams currently replaces öäü with oau so the information is lost before the indexer can handle it. From my observations this pr improves the results significantly. Are you suggesting that aio should use the unaltered umlaut containing title for querys to usenet indexers because they can handle it better? That would be an additional discussion.

Perhaps I misunderstood what this PR actually does. I thought the purpose was to implement the replacing of umlauts, I didn't know it already did that.
It's a double-edged sword. If you use NZBHydra2 you will notice it has an optional toggle to replace umlauts and diacritics, yet in my experience I've found that in some cases you will actually have fewer results. 
Most usenet indexers handle this intelligently by matching umlauts to their equivalents internally. Yet for those that do not, a normalized title that originally contained an umlaut (e.g. "Gefährliche Liebe" -> "Gefahrliche Liebe" or "Gefaehrliche Liebe") would no longer match a file named "Gefährliche Liebe" on the indexer's side.
In my experience this makes it worse for this specific case, but I don't know what implications this has for torrent indexers and other sources. There will be trade-offs either way.

nexariton:
I can't confirm this. I get by far the best results with this naming convention (ae oe ue) across the board.
Regardless this pr improves the current situation.
If you want to continue this discussion though maybe you could reference your last comment in a new feature request issue.

be-na17:
> I can't confirm this. I get by far the best results with this naming convention (ae oe ue) across the board. Regardless this pr improves the current situation. If you want to continue this discussion though maybe you could reference your last comment in a new feature request issue.

Ah, I see. The purpose is to not have ÄÖÜ replaced with AOU but rather AeOeUe. Yes, I believe that would be a big improvement 👍

Viren070:
A search of `die buchse der pandora` (current behaviour) on knaben returns 6 results but `die buechse der pandora` (new behaviour) returns none. 

Although, a search of `Das weiße Band` does indeed return no results and would need to be converted to `Das weisse Band`

So it does improve results in some cases but can also have the opposite effect in others. Would you say its overall increased results for you?

nexariton:
Yes in my tests it often went up from 0-1 results to a few dozen.

be-na17:
The naming convention in the German scene is inconsistent, I'm afraid. And it would be up to each individual indexer and how fuzzy their search is (the issue I noted above). Would it not be possible to perform two queries for titles containing umlauts when using text-based search? E.g. like the Force Query Search option for Newznab when querying for series.

Viren070 (approved):

hamburgercowgoldfish:
Can these ASCII equivalents also be accepted for title matching without lowering the threshold? I just had the case where results called "Doppelhaushaelfte" were filtered out for a title called "Doppelhaushälfte"

PR #483, opened 2025-11-10
feat: add separate size filter for anime content

### LINKED ISSUE #350: Feature Request: Add Anime type in Size filter
### Checklist

Many times when opening Anime episodes, there are these huge files:

Image

(sometimes even 100gb files or 300gb files)

that contains a full anime (or many seasons) in one file. Most of the time, you can't play them, and when you can they play the wrong episode (not the one you need).

### Proposed Solution

Add Anime type in Size filter:

Image

So we can set a Size limit for example of 15gb for Anime and avoid these huge useless packs

worldInColors:
- Add anime type to size filter settings
- Allow independent size limits for anime vs movies/series
- Prevents large anime pack files from appearing in results
- Resolves #350 

## Release Notes

* **New Features**
  * Added anime-specific size filtering alongside existing movie and series options.
  * Users can set independent global and resolution-specific size limits for anime.
  * UI now includes an Anime Size slider with min/max inputs and value display.
  * Filtering now applies anime-specific size constraints where applicable.


be-na17:
AIOStreams has started differentiating between anime movies and series internally since a recent update. Could this be further broken up into those two categories to create parity with the regular size filters?

Viren070:
> AIOStreams has started differentiating between anime movies and series internally since a recent update. Could this be further broken up into those two categories to create parity with the regular size filters?

You should just use stream expressions for this use case. I'd rather not make the UI filters more crowded than they already are. 

e.g

```
queryType == 'anime.series' ? size(streams, '2GB', 10GB') : []
```

in excluded stream expressions, will filter out streams between 2-10 GB for anime specifically, when its a series.

Viren070:
If we add this, it means that users wouldn't be able to set a global limit for movies including anime movies.

A user might set a size range of >500mb for anime, but this same range would override the range set for movies (e.g. if the user had also set a movie range for >2gb). Meaning - they would see files less than 2GB for anime movies, but this may not be what they want.

Now they'd have to use SEL so they could specifically apply the 500mb for anime series,  

Perhaps we could limit the anime UI filter to series only or we could add a switch in the size filter menu e.g. `Separate anime ranges`, and turning this on will add 2 ranges (one for anime movie, and one for anime series), as @ben-nasi suggested earlier. 

I think limiting to series would certainly be the easier option and most users would probably only want to have a separate size slider for (anime) series and not for movies. For those who do want a separate anime movie range - they could then use SEL.

This would also break existing configs for users who have a movie/size filter set - but because they don't have an anime size filter set, their movie/size ranges that usually would've worked will no longer work for anime. I believe this can be fixed by writing this section like this instead:

```js
      let minMax: [number | undefined, number | undefined] | undefined;
      if (type === 'movie') {
        minMax =
          (isAnime ? normaliseSizeRange(resolution?.anime) || normaliseSizeRange(global?.anime) : undefined) ||
          normaliseSizeRange(resolution?.movies) ||
          normaliseSizeRange(global?.movies);
      } else {
        minMax =
          (isAnime ? normaliseSizeRange(resolution?.anime) || normaliseSizeRange(global?.anime) : undefined) ||
          normaliseSizeRange(resolution?.series) ||
          normaliseSizeRange(global?.series);
      }
```

and if we were to limit the anime size range to series, then:

```js
      let minMax: [number | undefined, number | undefined] | undefined;
      if (type === 'movie') {
        minMax =
          normaliseSizeRange(resolution?.movies) ||
          normaliseSizeRange(global?.movies);
      } else {
        minMax =
          (isAnime ? normaliseSizeRange(resolution?.anime) || normaliseSizeRange(global?.anime) : undefined) ||
          normaliseSizeRange(resolution?.series) ||
          normaliseSizeRange(global?.series);
      }
```

be-na17:
I'd prefer the option with two sets of sliders, but admittedly I do have a horse in the race 😅
Most people will simply not bother with SEL, so in my opinion the more customization options they have over the UI, the better. An optional toggle for displaying anime sliders sounds like the most comprehensive solution to me.
If that is not possible, then perhaps simply having the movies/series sliders apply to anime.movies and anime.series, like they used to, would be more accessible than SEL to users.

Viren070 (approved):

PR #545, opened 2025-12-08
update package

PatrickMatthiesen:
A from what i can see, this fixes the NEXT JS CVE-2025-66478, caused by the upstream CVE-2025-55182.
I would also recommend that React and react-dom is updated to 19.2.1 as a minimum.

## Summary by CodeRabbit

* **Chores**
  * Updated frontend framework and React to newer patch/minor releases to bring in recent improvements and fixes.
  * Bumped type definitions and linting presets to align with updated framework versions and ensure consistent development/IDE experience.


Viren070:
updated packages in 7c5ea8a58c61dd8cc5f80031cb2769e8c9ede728 and e044139db6a839d61a38c9829fbc39ca3fecb327, though I believe this vulnerability doesn't apply to AIOStreams since we use the static export mode of next js which only produces pure html/css/js which is then served via express js and not using react server.

Updated to be safe anyways though.

PatrickMatthiesen:
Great, happy to see it, thanks a lot!

I also wasn't sure if we were affected, but I would rather be safe, as many people might not be running in a closed environment.

PR #553, opened 2025-12-15
FEAT: use content path instead of history for checking if usenet NZB is in library

mindgam3s:
FEAT: use content path instead of history for checking if usenet NZB is in library

## Release Notes

* **Bug Fixes**
  * Improved video format detection (HDR, DV, and variants) in content metadata parsing
  * Enhanced stream deduplication to distinguish between cached and uncached sources
  * Optimised NZB discovery through alternative source integration

* **Improvements**
  * Refined language detection for German content


mindgam3s:
@Viren070 could you please have a look at this?
I tested it myself by building the container locally and ever since using it like this.
would be great if you could merge this into the nightly!

Issue #580, opened 2025-12-29
feature request: disable stream cache for library add-ons


webstreamr:
### Checklist

- [x] I checked that this feature has not been requested before
- [x] I checked that this feature is not in the "Not planned" list
- [x] This feature will benefit the majority of users

### Problem Description / Use Case

I'd like to use a higher `STREAM_CACHE_TTL` but the only thing limiting me atm is the fact that library add-ons (only StremThru Store atm) are then cached as well which I want to avoid to get fresh results of manually added content.

### Proposed Solution

Currently thinking about one of the following
- introduce smth like `FORCE_STREMTHRU_STORE_STREAM_CACHE_TTL` which allows to override the global STREAM_CACHE_TTL. open question: only implement it for ST Store or for all add-ons. the latter feels like an overkill
- introduce smth like `DEFAULT_STREMTHRU_STORE_STREAM_CACHE_TTL` and add an option for StremThru Store to customize it. what about other add-ons? feels like an overkill to me in general
- introduce smth like `FORCE_DISABLE_LIBRARY_STREAM_CACHE` which then circumvents the stream cache for all add-ons / presets where `library: true` is set
- just disable stream cache in general for add-ons where `library: true` is set

I currently favor the last idea because it's the simplest and might help others too

Viren070:
Should be possible with 47eabefd64308ea45955311325a187d734cc6b9e

e.g.

```
STREAM_CACHE_TTL=*:300,stremthruStore:-1
```

webstreamr:
@Viren070 I just had a case with an uncached usenet (scenenzbs) result using torbox which did not update its uncached status even after it was already cached on torbox. can there be a side-effect by the stream cache introduced here? I thought uncached streams will always be checked again independently.

and are timeouts of direct add-ons cached as well? e.g. torbox usenet search timed out after 30s, after a retry I immediately got the timeout again without a delay.

sorry, I was assuming the stream cache sits before these things, but it seems it is after?

PR #581, opened 2025-12-29
feat: disable stream cache for library add-ons

webstreamr:
Closes https://github.com/Viren070/AIOStreams/issues/580

Simplest way of dealing with that, check proposed solutions there for possible alternatives.

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified stream caching behaviour for library addons to improve performance and resource management. Stream caching is now disabled for library addons, whilst remaining active for standard addons.

### DIFF
--- a/packages/core/src/wrapper.ts
+++ b/packages/core/src/wrapper.ts
@@ -213,7 +213,7 @@ export class Wrapper {
       { type, id },
       this.addon.timeout,
       validator,
-      Env.STREAM_CACHE_TTL != -1 ? streamsCache : undefined,
+      Env.STREAM_CACHE_TTL != -1 && !this.addon.library ? streamsCache : undefined,
       Env.STREAM_CACHE_TTL,
       this.preset.getCacheKey({
         resource: 'stream',


webstreamr:
Ah, I just had another idea: introduce an env var for library stream TTL which is also -1 by default. WDYT?

webstreamr:
@Viren070 sorry for the ping but what do you think of this

> Ah, I just had another idea: introduce an env var for library stream TTL which is also -1 by default. WDYT?

was thinking of just adding another env var e.g. `LIBRARY_STREAM_CACHE_TTL` or `STREAM_LIBRARY_CACHE_TTL` next to the already existing `STREAM_CACHE_TTL`. feels less drastic than the change here, but basically nobody benefits from it then :D what about the change here, or does it not make sense?

Viren070:
I'm not really fond of adding an env var specifically to control cache for library addons - I would prefer to have a much more customisable environment variable that allows you to control cache_ttls for any addon, perhaps using hostnames which would satisfy this use case as well as many others.

webstreamr:
> I'm not really fond of adding an env var specifically to control cache for library addons - I would prefer to have a much more customisable environment variable that allows you to control cache_ttls for any addon, perhaps using hostnames which would satisfy this use case as well as many others.

interesting idea. the only problematic thing for exactly my use case would be to differentiate st torz from st store I guess 🙈

Viren070:
Ah right, i guess we could also allow specifying preset ids.

so e.g. the value of this supposed environment variable could be:

`aiometadata.domain.com:-1,stremthruStore:-1`

Although I guess if theres another host which serves multiple addons and neither are available in the marketplace this variable wouldn't work for them, still I think this would work for everyone and still keep the variable simple.

webstreamr:
Sorry that I didn't find any time yet to further look into this and thank you for finishing it!

Viren070:
Not a problem at all, its me who wanted to do something different anyways 😅

Let me know if the feature works for you; I replied in your original issue for your specific use-case and detailed documentation was added to the [.env.sample](https://github.com/Viren070/AIOStreams/blob/f2f3232a44ddba57046d24c719994d4f7ffb66b0/.env.sample#L380-L400).

PR #605, opened 2026-01-04
feat: add cleanup after resolve functionality

### LINKED ISSUE #602: feature request: Option to not add streams to derbid dashboard
### Checklist

- [ ] This feature will benefit the majority of users

There are multiple use cases that this could benefit users, however this may benefit users of some services more than others.

My usecase is specifically from a Torbox perspectived

**Current implementation**

When a cached stream is chosen via AIOstreams, the watched item shows up on the Dashboard, as if added to library.  There are downsides to this:

1) If the Debrid is shared with multiple users, everything that everyone watches is added to the dashboard.  If multiple users are watching the same TV Shows for example, I could receive a library result that someone else has added by virtue of having watching it.  It is no longer safe to assume that a library item is indeed favourable.

2) When I play an episode from a season pack, the entire season is added to my dashboard.  If that then turns out to  be a Russian only file, I have to manually purge this from my dashboard or it comes up as a library item

3) I would prefer that Library items found via search are only because I have manually added something to the cache / dashboard.

4) This one is a bit of an outlier and definitely torbox specific.  I often have to manually add files to my dashboard because they have bad metadata.  They cannot be searched or launched from AIOStreams addons via search because for example, episodes don't match the files.  The only way to play these is to play them from Torbox "Your media"  However, this gets conjested with everything on the dashboard (everything everyone sharing my account is viewing) so it gets unweildy really quickly.

Currently I am having to regularly, manually purge my dashboard to make use of it the way I would like.

### Proposed Solution

Add an option per service, whether you would like to prevent streams selected by AIOStreams to be added to that service's dashboard.

This could either be that:

1) it never gets added
or
2) it is added and subsequently removed. 

Whichever is more technically feasible.

worldInColors:
closes #602 

## Summary by CodeRabbit

* **New Features**
  * Added "Auto remove Downloads" pro-mode setting to toggle automatic removal of cached downloads after playback.

* **Chores**
  * Setting is respected across core, TorBox search and presets so sources honour the choice.
  * Optional background cleanup for both torrent and usenet downloads runs after playback and logs failures without blocking playback.

### DIFF
--- a/packages/frontend/src/components/menu/miscellaneous.tsx
+++ b/packages/frontend/src/components/menu/miscellaneous.tsx
@@ -307,7 +307,7 @@ function Content() {
             <Switch
               label="Enable"
               side="right"
-              value={userData.autoRemoveDownloads}
+              value={userData.autoRemoveDownloads ?? false}
               onValueChange={(value) => {
                 setUserData((prev) => ({
                   ...prev,


DanaramaPyjama:
Thanks for looking into this chaps

packages/core/src/builtins/base/debrid.ts
```diff
@@ -254,7 +255,8 @@ export abstract class BaseDebridAddon<T extends BaseDebridConfig> {
         torrentServices,
         id,
         searchMetadata,
-        this.clientIp
+        this.clientIp,
+        this.userData.autoRemoveDownloads
       ),
```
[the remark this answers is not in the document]
worldInColors: The solution is to remove the parameter, not implement it

worldInColors:
Accidentally removed that while i was removing autoRemoveDownloads

packages/core/src/debrid/torbox.ts
```diff
@@ -49,6 +49,33 @@ export class TorboxDebridService implements DebridService {
     return this.stremthru.listMagnets();
   }
 
+  public async removeMagnet(magnetId: string): Promise<void> {
+    return this.stremthru.removeMagnet(magnetId);
+  }
+
+  public async removeNzb(nzbId: string): Promise<void> {
+    try {
+      await this.torboxApi.usenet.controlUsenetDownload(this.apiVersion, {
```
Viren070, line 58: You need to check the return value of controlUsenetDownload for status / success / errors and throw DebridError when appropriate, it might return a value in `result.data?.success`, if this is false, error needs to be checked in result.data?.error or result.data?.detail. the sdk returns http status errors in result.metadata.status and result.metadata.statusText which can be used when throwing the DebridError. 

check what is returned from this function and add appropriate handling.

packages/frontend/src/components/menu/miscellaneous.tsx
```diff
@@ -284,6 +284,39 @@ function Content() {
             />
           </SettingsCard>
         )}
+        {mode === 'pro' && (
+          <SettingsCard
+            title="Auto remove Downloads"
+            description={
+              <div className="space-y-2">
+                <p>
+                  When enabled, debrid services will automatically remove the
```
Viren070:
```suggestion
                  When enabled, AIOStreams will automatically remove the
```

Viren070 (approved):

DanaramaPyjama:
:heart:  :heart:  :heart:

Issue #641, opened 2026-01-23
feature request: set private flag for torznab indexer results accordingly


webstreamr:
### Checklist

- [x] I checked that this feature has not been requested before
- [x] I checked that this feature is not in the "Not planned" list
- [ ] This feature will benefit the majority of users

### Problem Description / Use Case

Currently, when private torrent indexers are connected via e.g. jackett torznab, their private flag is not set properly and therefore using it in the formatter makes no sense. Would be great if this can be set properly for such cases as well. For results from ST Torz it seems to work already.

### Proposed Solution

Described above already, I'm just not sure if that info is already there in the torznab results or not and how complicated it is to use it. If you point me to the right places I can also try to implement it.

webstreamr:
hmm I'm already digging a bit and only saw so far that the info seems to be not there in e.g. debrid getStreams() :/

update: was digging closer and hooking into the torznab result where I saw the following

```json
{
  "title": "28.Weeks.Later.2007.German.DTS.DL.1080p.BluRay.x264-VECTOR.mkv",
  "link": "<REDACTED>",
  "guid": "<REDACTED>",
  "pubDate": "Thu, 27 Nov 2025 22:56:24 +0000",
  "jackettindexer": {
    "name": "RocketHD (API)",
    "id": "rockethd"
  },
  "size": 10931432448,
  "enclosure": [
    {
      "url": "<REDACTED>",
      "length": 10931432448,
      "type": "application/x-bittorrent"
    }
  ],
  "torznab": {
    "category": 100001,
    "imdb": 463854,
    "imdbid": "tt0463854",
    "tmdbid": 1562,
    "genre": "Horror, Thriller, Science Fiction",
    "seeders": 7,
    "peers": 7,
    "coverurl": "<REDACTED>",
    "minimumseedtime": 604800,
    "downloadvolumefactor": 1,
    "uploadvolumefactor": 1
  }
}
```

doesn't look useful :/

update2: Munif enlightened me :D
> If you download the torrent file and parse it, there will be a private flag.

webstreamr:
got something relatively simple working, but might still need some cleanups. I'll push it a bit later and open a PR :)

PR #642, opened 2026-01-23
chore(formatters): add private flag

webstreamr:
mostly an idea to start the discussion. I think marking private results is useful, but not sure how users will react if they see this suddenly without configuring any private indexers even (e.g. ST Torz). Also not sure what the best symbol is, the lock is obviously already take. Torz itself is using the key.

Combined with https://github.com/Viren070/AIOStreams/issues/641 it would be very useful for people connecting private indexers.

## Release Notes

* **Enhancements**
  * Added a visual indicator (🔑) to display when streams are marked as private
  * Improved consistency across stream formatters in how stream metadata is presented
  * Refined the ordering and alignment of stream property indicators for enhanced clarity

### DIFF
--- a/packages/core/src/formatters/predefined.ts
+++ b/packages/core/src/formatters/predefined.ts
@@ -6,7 +6,7 @@ export class TorrentioFormatter extends BaseFormatter {
     super(
       {
         name: `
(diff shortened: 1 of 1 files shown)


Viren070:
I think 🔑 is also the best option to use here

PR #644, opened 2026-01-23
feat(torznab): make use of torrent private flag

webstreamr:
Closes https://github.com/Viren070/AIOStreams/issues/641

~~This works, but feels stupid because it has to set that property also in the stream to get the info into the parsed stream. is there a better way?~~ found a slightly better way that does not misuse the stream object

image

## Release Notes

* **New Features**
  * Private torrents are now identified and visually marked with a 🔑 lock indicator in stream displays.
  * Comprehensive private status tracking has been integrated across torrent metadata, enrichment, and playback processing.
  * Enhanced stream parsing to automatically detect and mark private content for improved visibility.

### DIFF
--- a/packages/core/src/presets/torznab.ts
+++ b/packages/core/src/presets/torznab.ts
@@ -1,24 +1,10 @@
-import { Addon, Option, ParsedStream, Stream, UserData } from '../db/index.js';
+import { Addon, Option, Stream, UserData } from '../db/index.js';
 import { Preset, baseOptions } from './preset.js';
 import { Env, RESOURCES, ServiceId, constants } from '../utils/index.js';
 import { StremThruPreset } from './stremthru.js';
 import { BuiltinAddonPreset } from './builtin.js';
-import { StreamParser } from '../parser';
-
-export class TorznabStreamParser extends StreamParser {
-  protected override isPrivate(
-    stream: Stream,
-    _currentParsedStream: ParsedStream
-  ): boolean | undefined {
-    return stream.name?.includes('🔑') ? true : false;
-  }
-}
 
 export class TorznabPreset extends BuiltinAddonPreset {
-  static override getParser(): typeof StreamParser {
-    return TorznabStreamParser;
-  }
-
   static override get METADATA() {
     const supportedResources = [constants.STREAM_RESOURCE];
     const options: Option[] = [
--- a/packages/core/src/presets/builtin.ts
+++ b/packages/core/src/presets/builtin.ts
@@ -71,6 +71,13 @@ export class BuiltinStreamParser extends StreamParser {
     }
   }
 
+  protected override isPrivate(
+    stream: Stream,
+    _currentParsedStream: ParsedStream
+  ): boolean | undefined {
+    return stream.name?.includes('🔑') ? true : false;
+  }
+
   protected getStreamType(
     stream: Stream,
     service: ParsedStream['service'],


packages/core/src/utils/torrent.ts
```diff
@@ -196,6 +199,7 @@ export class TorrentClient {
           })
         ),
         sources,
+        private: !!parsedTorrent.info?.private,
```
webstreamr, line 202: the type is boolean but still I got `1` there..

packages/core/src/debrid/stremthru.ts
```diff
@@ -329,6 +329,7 @@ export class StremThruInterface implements DebridService {
       size: magnetDownload.size || 0,
       type: 'torrent',
       sources: playbackInfo.sources,
+      private: false,
```
webstreamr: not sure when this stremthru implementation is used and if it can be also `true` here

packages/core/src/builtins/torbox-search/source-handlers.ts
```diff
@@ -340,10 +340,13 @@ export class TorrentSourceHandler extends SourceHandler {
...
     const { results, errors } = await processTorrents(
-      fetchResult.torrents.map((torrent) => ({
-        ...torrent,
-        type: 'torrent',
-      })),
+      fetchResult.torrents.map((torrent) => {
+        return {
+          ...torrent,
+          type: 'torrent',
+          private: false,
```
webstreamr, line 348: not sure if torbox search might return `true` here. didn't find anything in the api

packages/core/src/builtins/base/debrid.ts
```diff
@@ -792,6 +793,7 @@ export abstract class BaseDebridAddon<T extends BaseDebridConfig> {
       duration: torrentOrNzb.duration,
       infoHash: torrentOrNzb.hash,
       fileIdx: torrentOrNzb.file.index,
+      private: torrentOrNzb.type === 'torrent' ? torrentOrNzb.private : undefined,
```
webstreamr: this is the change I don't like. having this flag here is false, right?

webstreamr:
@Viren070 the changes here are adapted to not misuse the `Stream` object and they are basically scoped to torznab only now. do you think this goes in the right direction?

packages/core/src/builtins/torbox-search/source-handlers.ts
Viren070: yes pretty sure they dont.

packages/core/src/debrid/stremthru.ts
Viren070: Per the sdk, magnetDownload.magnet is available here from StremThru, we may need to update the SDK version if its not up to date. 

The value of it wouldn't really matter though as it doesnt affect anything, but I think making it "correct" would be a good idea regardless.

webstreamr: hmm if this is magnet-only code then it can stay false I think.

webstreamr: just saw that I can pass-through the info to have it correct here and pushed a change. wdyt? I didn't get the point about the SDK tbh, but feel free to give me any pointers

packages/core/src/debrid/base.ts
```diff
@@ -126,6 +126,7 @@ const BaseFileInfoSchema = z.object({
 
 const TorrentInfoSchema = BaseFileInfoSchema.extend({
   hash: z.string(),
+  private: z.boolean(),
   sources: z.array(z.string()),
   type: z.literal('torrent'),
 });
```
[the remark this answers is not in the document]
webstreamr: ah is that true Viren and does the default false suggestion make sense? 🤔

packages/core/src/presets/torznab.ts
Viren070, line 1: Rather than adding a TorznabStreamParser, we can extend the existing BuiltinStreamParser and override `isPrivate` there, This way it applies to other built-ins like Prowlarr and perhaps other torrent addons in the future, 

Also, even if we were to do it this way, we would need the `TorznabStreamParser` to extend the `BuiltinStreamParser` as that likely has other logic to parse other attributes; removing that and only having the `isPrivate` override might cause unexpected behaviour.

webstreamr: good catch, adapted

PR #664, opened 2026-01-28
feat(debrid): implement adding torrent via download URL

webstreamr:
passes-through `downloadUrl` and implements `addTorrent()` analogue to `addMagnet()` using stremthru store

first private torrent I added using this :)

image

the only thing that worries me a bit, and it might be a big thing actually, what if people use their internal jackett or prowlarr (via localhost:xxxx) but an external stremthru or their stremthru is not on the same host? in such cases stremthru would not be able to download the torrent I guess. maybe it's a self-hosting special case, not sure. alternative could also be, but that would be more work obviously, that aiostreams downloads the torrent and passes the data to stremthru. 

## Summary by CodeRabbit

* **New Features**
  * Debrid services now support adding torrents via direct download URLs and include a downloadUrl field in torrent file/playback info.
  * Multiple debrid backends expose an add‑torrent API (some implementations initially stubbed).

* **Chores**
  * Updated core dependency for stremthru to the latest minor version.

### DIFF
--- a/packages/core/src/debrid/stremthru.ts
+++ b/packages/core/src/debrid/stremthru.ts
@@ -278,7 +278,7 @@ export class StremThruInterface implements DebridService {
     }
 
     let magnetDownload: DebridDownload;
-    if (playbackInfo.downloadUrl) {
+    if (playbackInfo.downloadUrl && Env.BUILTIN_DEBRID_USE_TORRENT_DOWNLOAD_URL) {
       logger.debug(
         `Adding torrent to ${this.serviceName} for ${playbackInfo.downloadUrl}`
       );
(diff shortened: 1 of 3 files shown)


packages/core/src/debrid/stremthru.ts
```diff
@@ -276,14 +299,36 @@ export class StremThruInterface implements DebridService {
...
+
+      logger.debug(`Adding magnet to ${this.serviceName} for ${magnet}`);
+
+      magnetDownload = await this.addMagnet(magnet);
+
+      logger.debug(`Magnet download added for ${magnet}`, {
+        status: magnetDownload.status,
+        id: magnetDownload.id,
+      });
+    }
```
webstreamr, line 309: should not be an issue. and if it's changed, then the existing magnet logic would need to be kept in sync.

webstreamr:
added env var to opt-out and revert back to the magnet-only behaviour for self-hosted setups with internal jackett/prowlarr but without self-hosted stremthru basically.

Viren070 (approved):

PR #665, opened 2026-01-29
Fix: Skip bitrate calculation for incorrect metadata

Crunging:
When metadata is incorrect (has a placeholder runtime of 1), or there is a mismatch between seasons and episodes, skip the bitrate calculation.

This primarily happening when users are using AIOM for metadata, but are not setting their TVDB key in AIOS settings. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Bitrate calculation now skips special releases (Season 0) to prevent incorrect adjustments.
  * Added clearer debug logging when bitrate computation is bypassed for inapplicable episodes, reducing misleading diagnostics.

### DIFF
--- a/packages/core/src/streams/filterer.ts
+++ b/packages/core/src/streams/filterer.ts
@@ -315,38 +315,6 @@ class StreamFilterer {
           );
         }
 
-        // checks for metadata mismatches
-        if (
-          type === 'series' &&
-          requestedMetadata?.seasons &&
-          stream.parsedFile?.seasons?.length
-        ) {
-          const metadataSeasons = requestedMetadata.seasons;
-
-          for (const seasonNum of stream.parsedFile.seasons) {
-            const seasonData = metadataSeasons.find(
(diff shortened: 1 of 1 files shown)


packages/core/src/streams/filterer.ts
```diff
@@ -303,6 +303,40 @@ class StreamFilterer {
...
+        ) {
+          const metadataSeasons = requestedMetadata.seasons;
+
+          for (const seasonNum of stream.parsedFile.seasons) {
+            const seasonData = metadataSeasons.find(
+              (s) => s.season_number === seasonNum
+            );
+
+            if (!seasonData) {
+              if (!this.userData.tvdbApiKey) {
```
Crunging: This if statement is kind of a bandaid fix right now as tvdb does not provide season metadata in service.ts and tvdb.ts. I can of course try and implement it if you'd like, I was just focused on a small quick fix.

PR #666, opened 2026-01-29
feat: added EZTV as an builtin addon, uses EZTVs API to fetch torrents

bytedeal3r:
* Added EZTV builtin addon support.
* Uses EZTVs official API.
* The API is in BETA, but works most of the time. (saw some episodes not being fetched on some shows)
* Implemented the same way Knaben & TorrentGalaxy was, with some modifications to fit the EZTV response object.

## Summary by CodeRabbit

* **New Features**
  * Added EZTV as a built‑in debrid source for TV series with integrated torrent search, deduplication and manifest/stream JSON endpoints.
  * New EZTV preset with UI options and manifest URL generation.

* **Chores**
  * Configurable EZTV URL, timeouts and search cache TTL via environment settings.
  * Logging label added for EZTV.

### DIFF
--- a/packages/core/src/builtins/eztv/api.ts
+++ b/packages/core/src/builtins/eztv/api.ts
@@ -168,13 +168,13 @@ class EztvAPI {
         timeout: options.timeout ?? Env.MAX_TIMEOUT,
       });
 
-      const data = (await response.json()) as unknown;
-
       if (!response.ok) {
         throw new Error(
           `EZTV API error (${response.status}): ${response.statusText}`
         );
       }
+
+      const data = (await response.json()) as unknown;
 
       try {
         return schema.parse(data);


packages/core/src/builtins/eztv/api.ts
```diff
@@ -0,0 +1,179 @@
...
+  private readonly searchCache = Cache.getInstance<
+    string,
+    EztvGetTorrentsResponse
+  >('eztv:search');
+
+  constructor() {
+    this.headers = {
+      'Content-Type': 'application/json',
+      'User-Agent':
+        'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36',
```
Viren070: is there a reason a browser UA is being used here instead of Env.DEFAULT_USER_AGENT? 

if so, maybe we should add Env.BUILTIN_EZTV_USER_AGENT which defaults to this if there is a reason to use a browser UA here instead of the default

packages/core/src/builtins/eztv/api.ts
```diff
@@ -0,0 +1,179 @@
...
+  DistributedLock,
+} from '../../utils/index.js';
+import { createLogger } from '../../utils/index.js';
+import { searchWithBackgroundRefresh } from '../utils/general.js';
+
+import { z } from 'zod';
+
+const logger = createLogger('eztv');
+
+const EztvTorrentSchema = z.object({
```
Viren070: this schema should have a transform added to it to match other schemas for builtin addons. mapping attribute names to camelCase

Viren070 (changes requested):

packages/core/src/builtins/eztv/api.ts
bytedeal3r: no reason, just saxed the implementation from TGx & Knaben, which used this user agent.

packages/core/src/builtins/eztv/api.ts
bytedeal3r: fixed :)

PR #681, opened 2026-02-04
FEAT edge-case for DVHDR and DVHDR10(+) etc

mindgam3s:
in some streams the DV and HDR is not separated by a dot but instead is bundled together like so: filename.DVHDR10p.extension

this solves it by allowing this for both DV and HDR tags to be a valid match

## Summary by CodeRabbit

* **Improvements**
  * Improved recognition of HDR-related video tags: Dolby Vision-prefixed variants and HDR-influenced Dolby Vision naming are now detected more reliably, enhancing compatibility with diverse video metadata and catalogue formats.


mindgam3s:
@Viren070 
same movie

before:
before
after:
after

PR #688, opened 2026-02-05
FEAT edge-case for DVHDR and DVHDR10(+)

mindgam3s:
in some streams the DV and HDR is not separated by a dot but instead is bundled together like so: filename.DVHDR10p.extension
this solves it by allowing this for both DV and HDR tags to be a valid match

@Viren070 
same movie

before:
before
after:
after

## Summary by CodeRabbit

**Bug Fixes**
- Enhanced video metadata tag recognition to better support various naming conventions for HDR formats, Dolby Vision variants, and Atmos audio options. Improved compatibility with diverse file naming standards results in more accurate metadata extraction and recognition across a wider range of video formats and codecs.


hugo9655:
This is not the right way to approach this problem imo.

You could easily do the same for any two that someone has placed together, and then the patterns will become so complex it'll look silly.

The only other way (which I also dont recommend) would be to remove word boundary checks (which seems to be implemented separate from the regex patterns themselves) but that will impact other filters negatively and cause false positives.

There will always be people who format releases horribly, some cases just can't easily be addressed unfortunately.

mindgam3s:
> This is not the right way to approach this problem imo.
> There will always be people who format releases horribly, some cases just can't easily be addressed unfortunately.

to be honest this is quite a common naming scheme for german releases...
indicating that this is `DV with HDR(10)(plus) fallback`
that's how it got my attention that some streams where "incorrectly" tagged by AIOStreams.

> You could easily do the same for any two that someone has placed together, and then the patterns will become so complex it'll look silly.

and I didn't create a pull request for other bundled-together tags, it was specifically for this case that happens a lot of the time in german releases.


> The only other way (which I also dont recommend) would be to remove word boundary checks (which seems to be implemented separate from the regex patterns themselves) but that will impact other filters negatively and cause false positives.

which I would agree to not do. that's why I added the pattern above for this one specific case `DV with HDR(10)(plus) fallback`

hugo9655:
> and I didn't create a pull request for other bundled-together tags, it was specifically for this case that happens a lot of the time in german releases.

It sets a precedent, that is not good.

mindgam3s:
> It sets a precedent, that is not good.

cannot really agree with that, as I said it is not uncommon for german release groups to name it like this

(one of the biggest to do this as well is **pmHD** which is **German Web Tier 01**)

PR #692, opened 2026-02-06
feat: add toggle all on/off button for addons

### LINKED ISSUE #659: feature request: buttons to toggle all addons on/off
### Checklist

When experimenting with a growing list of addons, it can be tedious to individually toggle all addons on or off. Sometimes I want to quickly turn off all addons, and then enable just the one I am tinkering with.

This would be somewhat similar to the AIOMetadata Catalogs view where you can "Select all" and then apply an action to the selection. I think for AIOStreams, just toggling on/off would be enough.

### Proposed Solution

In the My Addons UI, above the list, add two buttons. One to toggle all addons off, and the other to toggle them all on.

evanesce345:
Wasn't sure the best UI/UX to go with for this, so it can be changed if needed.

Current implementation has a single button and is easy to understand:
image

Another idea is separate buttons which looks cleaner but is less intuitive:
image

Closes #659

## Release Notes

* **New Features**
  * Added a "Toggle all" switch to the My Addons settings, enabling users to quickly enable or disable all presets simultaneously.

### DIFF
--- a/packages/frontend/src/components/menu/addons.tsx
+++ b/packages/frontend/src/components/menu/addons.tsx
@@ -378,6 +378,24 @@ function Content() {
             <SettingsCard
               title="My Addons"
               description="Edit, remove, and reorder your installed addons. If you reorder your addons, you will have to refresh the catalogs if you have made any changes, and also reinstall the addon."
+              action={
+                <Switch
+                  id="toggle-addons"
+                  label="Toggle all"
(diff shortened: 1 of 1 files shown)


Viren070:
I'm closing this in favour of the redesigned my addons section added in 4652ef1314ed1a426dfa24887e298ad29741f2ea which allows selecting and toggling addons.

PR #693, opened 2026-02-06
chore(frontend): replace "magnet" with "torrent" in auto remove downloads

webstreamr:
To be a bit clearer that this is not limited to magnets and will affect e.g. private torrents added without magnets as well. but most likely it only confused me with magnet/torrent internal info..

## Summary by CodeRabbit

* **Documentation**
  * Updated the description text for the Auto remove Downloads feature to accurately reflect "torrent/NZB" instead of "magnet/NZB".


webstreamr:
superseded by https://github.com/Viren070/AIOStreams/pull/694

PR #729, opened 2026-02-13
fix(debrid): limit torbox magnet checks to 2 per user at any given time

webstreamr:
this is a simple and stupid workaround to avoid hitting the 5/s limits with the current aiostreams design. it seems to work quite well on my local system and should still leave room for wrapped add-ons doing their own availability checks.

in case this gets merged - something similar will be needed for nzb checks too I suppose. I only have a single indexer and don't experience problems there though.

## Summary by CodeRabbit

* **Dependencies**
  * Added lru-cache dependency to manage cached controls and improve stability.

* **Performance Improvements**
  * Improved magnet validation by introducing per-account concurrency limits, reducing request spikes and improving overall responsiveness and resource efficiency.

### DIFF
--- a/packages/core/src/debrid/torbox.ts
+++ b/packages/core/src/debrid/torbox.ts
@@ -17,6 +17,8 @@ import {
 } from './base.js';
 import { StremThruInterface } from './stremthru.js';
 import { ParsedResult, parseTorrentTitle } from '@viren070/parse-torrent-title';
+import pLimit, { LimitFunction } from 'p-limit';
+import { LRUCache } from 'lru-cache';
 
 const logger = createLogger('debrid:torbox');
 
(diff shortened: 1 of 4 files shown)


packages/core/src/debrid/torbox.ts
```diff
@@ -148,7 +154,13 @@ export class TorboxDebridService implements DebridService {
   }
 
   public async checkMagnets(magnets: string[], sid?: string) {
-    return this.stremthru.checkMagnets(magnets, sid);
+    let limiter = TorboxDebridService.checkMagnetsLimiters.get(this.token);
+    if (!limiter) {
+      limiter = pLimit(2);
```
webstreamr, line 159: not sure what a good number here is tbh and it's a bit hard to test right now because some systems have intermittent issues. maybe it should just be 1 🤷‍♂️

webstreamr:
TorBox raised limits to 300/s and ST switched to non-batched post requests. This dirty workaround is not needed for now.

PR #730, opened 2026-02-13
fix(debrid): run torbox usenet availibility batch checks in sequence

webstreamr:
first step to avoid 5/s errors in this endpoint. if requests are done in batches because they are more than 100, then they will run in sequence instead of parallel per add-on

## Release Notes

**Bug Fixes**
- Improved error handling and validation for Torbox USENET cached availability checks. Error messages are now more descriptive and informative, including comprehensive details about failures. This enhancement increases visibility into issues when they occur, allowing for faster issue resolution and strengthening the overall reliability of USENET cache checking on Torbox.

### DIFF
--- a/packages/core/src/debrid/torbox.ts
+++ b/packages/core/src/debrid/torbox.ts
@@ -196,42 +196,41 @@ export class TorboxDebridService implements DebridService {
       }
 
       try {
-        const batchResults = await Promise.all(
-          batches.map(async (batch) => {
-            const result =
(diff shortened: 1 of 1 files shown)


webstreamr:
TorBox increased limits to 300/m which should be enough. Aiostreams should still switch to POST using more hashes, maybe 500 per call in sequence as ST does for torrents.

For that https://github.com/TorBox-App/torbox-sdk-js/issues/2 is needed

PR #731, opened 2026-02-13
fix: Stop overrides from mismatching in the backend

Crunging:
The fallback was causing an issue where if you only had an override for web T1 and nothing else, anime web T1 would first check name matches, then fallback to legacy name matches where it would then incorrectly apply the override. This override does not show in the frontend, causing mismatched buggy behavior. I think just getting rid of the fallback is the cleanest and most bug free way to deal with this.

## Summary by CodeRabbit

* **Refactor**
  * Refined selector override matching to use exact-name criteria only, removing legacy name variant matching from override resolution and streamlining the matching logic.


Crunging:
Updated description and title to more accurately describe the issue

PR #743, opened 2026-02-18
fix(parser): extend 10bit visual tag to match hi10 (in addition to hi10p)

Vidhin05:
## Summary
Extends the 10bit visual tag regex to also match `hi10` (in addition to the existing `hi10p` match). This covers torrents that use the shorter "hi10" tag for H.264 High 10 Profile content, commonly found in anime releases.

## Change
- **Before:** `10[ .\-_]?bit|hi10p` – matched only `hi10p`
- **After:** `10[ .\-_]?bit|hi10p?` – matches both `hi10` and `hi10p`

The `?` makes the trailing `p` optional, so both variants are recognized.

## Reference
- Regex tested with Seadex releases at: https://regex101.com/r/GsncCf/

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced 10-bit video format detection to recognise additional naming conventions, improving content classification accuracy.

### DIFF
--- a/packages/core/src/parser/regex.ts
+++ b/packages/core/src/parser/regex.ts
@@ -79,7 +79,7 @@ export const PARSE_REGEX: PARSE_REGEX = {
     SCR: createRegex('((dvd|bd|web|hd)?[ .\\-_]?)?(scr(eener)?)'),
   },
   visualTags: {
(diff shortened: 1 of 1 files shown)


emilyrichardssophie-web:
Why would you merge this??

PR #751, opened 2026-02-20
feat(presets/torznab): add configurable result limit

webstreamr:
surfaces in the UI what nzbhydra has already to force an initial limit. used without pagination it is a perfect way to reduce the amount of results in general for performance reasons.

not sure if this is useful for newznab as well and if it can become somewhat more generalised maybe.

nzbhydra has 250 set by default for it, but I went without a default. jackett seems to return limits.max=100 in general btw. 

## Summary by CodeRabbit

* **New Features**
  * Optional "Initial Result Limit" added for NAB and Torznab presets to set a per-request maximum number of search results.
  * Accepts values from 1 to 10,000; default remains unset/optional.
  * When specified, manifests and search responses will honour this cap to control result volume and reduce the chance of hitting external limits.

### DIFF
--- a/packages/core/src/presets/torznab.ts
+++ b/packages/core/src/presets/torznab.ts
@@ -118,7 +118,7 @@ export class TorznabPreset extends BuiltinAddonPreset {
         id: 'initialLimit',
         name: 'Initial Result Limit',
         description:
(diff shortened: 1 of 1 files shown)


webstreamr:
realized the original implementation was stupid and what I actually needed was just to expose a config for the already existing `forceInitialLimit`

PR #757, opened 2026-02-21
Get file size from NzbDav history instead of WebDAV

frogmech:
Get file size from NzbDav history instead of WebDAV so built-in library addon accurately reports file size

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-size accuracy in library and NZB listings by using recorded historical size as a fallback when immediate file size is unavailable or non-positive.
  * History entries now carry optional size metadata so listings display more consistent and reliable sizes when current file-size data is missing.


Viren070:
Is the size from webdav not accurate?

frogmech:
> Is the size from webdav not accurate? 

I probably could've worded that better, the library add-on doesn't seem to report any file size at all in the stream list, this PR fixes that.

PR #758, opened 2026-02-22
chore(presets): remove StreamAsia

webstreamr:
this seems to be dead now already for quite a while and nothing to be heard from the dev. was also never open source :/

https://stremio-dramacool-addon.xyz/

## Release Notes

* **Removals**
  * Removed the StreamAsia integration from the application, including the preset functionality, configuration options (URL, timeout, and user agent settings), and associated startup logging. Users relying on this integration should seek alternative solutions.


webstreamr:
was wondering how the UI behaves with this removed when people were using it. looks like it does not behave well 😭

image

Viren070:
We can add:

```js
      DISABLED: {
        reason: '',
        disabled: true,
        removed: true,
      },
```


 to the METADATA return value, the `removed` field will handle it so that it still shows the logo etc. on the users list but wont show in the marketplace and will error on save while still continue to work for existing users making normal stremio requests even if they have the addon in their list

webstreamr:
good idea, I wanted to do a drive-by cleanup and am too lazy to adapt that here and now. but I'll try to remember :D

PR #761, opened 2026-02-22
feat(builtins): add torznab freeleech info

webstreamr:
makes use of the `downloadvolumefactor` info in torznab results to determine if something is freeleech or not. via e.g. `{stream.freeleech::istrue[" FL "||""]} ` it is then possible to mark such streams in the formatter which is highly usable when private trackers are connected.

Refs: https://en.wikipedia.org/wiki/Glossary_of_BitTorrent_terms#Freeleech
Refs: https://torznab.github.io/spec-1.3-draft/revisions/1.2-Torznab-Download-Upload-Factors.html

image

## Summary by CodeRabbit

* **New Features**
  * Detect and label freeleech torrents so they’re clearly marked in stream names.
  * Support for Torznab-provided download volume factor to improve torrent handling.
  * Streams and stored torrent records now include a freeleech indicator for better categorisation and selection.

### DIFF
--- a/packages/core/src/builtins/base/debrid.ts
+++ b/packages/core/src/builtins/base/debrid.ts
@@ -866,8 +866,9 @@ export abstract class BaseDebridAddon<T extends BaseDebridConfig> {
         ? '⚡'
         : '⏳'
       : '';
+    const isFreeleech = torrentOrNzb?.downloadvolumefactor === 0;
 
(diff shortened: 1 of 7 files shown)


webstreamr:
Thank you for the merge. It's unfortunate that I can't contribute to the wiki, can you add this variable if you have a free moment please? Thx!

PR #793, opened 2026-03-08
feat(core/formatters): support variable resolution in replace() first argument

### LINKED ISSUE #790: feature request: Enhance replace method for formatters
### Checklist

The replace method, as of now, only work with _hardcoded_ strings

`filename`: Stranger Things
`{stream.filename::replace('St', '')}` results in: ranger Things

### Proposed Solution

It could be useful to make the replace method work also with all the available variables in order to enhance the quality of any  custom formatters

For example: 

`filename`: Stranger Things S1 E04
`title`: Stranger Things

`{stream.filename::replace('{stream.title}', '')}` results in: S1 E04

It could be useful, for example, to extract the episode name from the filename

kRYstall9:
This feature makes the replace method work with any formatter variable, such as {stream.title}, {addon.name} etc...
This might be useful for any person who wants to take their custom formatter any step forward.

### AS IS
`filename`: Movie.Title.2023.2160p.BluRay.HEVC.DV.TrueHD.Atmos.7.1.iTA.ENG-GROUP.mkv
`{stream.filename::replace('.', ' ')::replace('{stream.title}', '')}` replaces all the dots with a whitespace but it doesn't replace the stream title (Movie Title in this case) with an empty string

### TO  BE
`filename`: Movie.Title.2023.2160p.BluRay.HEVC.DV.TrueHD.Atmos.7.1.iTA.ENG-GROUP.mkv
`{stream.filename::replace('.', ' ')::replace('{stream.title}', '')}` replaces all the dots with a whitespace and it does replace the stream title (Movie Title in this case) with an empty string. 
So it will result in `2023 2160p BluRay HEVC DV TrueHD Atmos 7 1 iTA ENG-GROUP mkv`

Resolves #790 

## Summary by CodeRabbit

* **Improvements**
  * Replace operations now resolve dynamic placeholders at runtime before performing replacements; errors or empty resolutions preserve the original placeholder to avoid unintended broad matches.
  * Modifier application carries runtime value-resolution context through the formatting pipeline, allowing multi-modifier chains to operate on resolved values.
  * No breaking public API changes.

### DIFF
--- a/packages/core/src/formatters/base.ts
+++ b/packages/core/src/formatters/base.ts
@@ -896,7 +896,11 @@ export abstract class BaseFormatter {
             new RegExp(`${findStartChar}\\s*,\\s*${findEndChar}`)
           );
 
-          if (!shouldBeUndefined && key && replaceKey !== undefined) {
+          if (
+            shouldBeUndefined === undefined &&
+            key &&
+            replaceKey !== undefined
+          ) {
             let resolvedKey = key;
             if (key.startsWith('{') && key.endsWith('}') && parseValue) {
               // When the first argument to replace(...) is a {variable} expression, resolve it


kRYstall9:
Hi @Viren070!  I've addressed all the CodeRabbit suggestions. Could you take a look when you have a chance? Thanks!

PR #927, opened 2026-05-06
fix(config): skip precache selector validation when feature is disabled (#913)

### LINKED ISSUE #913: bug: SEL for "Pre-cache Next Episode" is Run Even When Disabled
### Deployment Method

VPS

### Addon Version

v2.28.0

### Bug Area

Obtaining streams.

### Device/Browser/OS/Stremio Version

_No response_

### Bug Description / Steps to Reproduce

`SILLY  2026-04-30 20:39:23.557 EDT | ✨ SEL | Expression evaluated in 0ms: "count(cached(streams)) == 0 ? uncached(streams) : []"`

This is in the log even when Misc - Background - PRE-CACHE NEXT EPISODE is disabled.

The SEL is the default from fresh config.

### Expected Behavior

If disabled, the SEL should be ignored. While the default SEL is very minor, a user could have typed in a more complicated one and disabled it, but it would still get run and wastes unnecessary step, especially in public instances.

### Configuration Export

Log shown in description.

### Screenshots

_No response_

### Debugging Checklist

- [ ] If applicable, I have included server logs 
- [ ] If applicable, I have included MediaFlow logs

### Issue Checklist

- [x] I have written a short but informative title that clearly describes the issue.
- [x] I have given clear and descriptive steps to reproduce the issue.
- [x] I have checked open and closed issues and confirmed that this is not a duplicate of another issue.
- [x] I have filled out all of the requested information adequately. 
- [x] I am using the [latest version](https://github.com/Viren070/AIOStreams/releases/latest).

Sanjays2402:
Closes #913.

## Diagnosis

`validateConfig` (in `packages/core/src/utils/config.ts`) is called from the userData middleware (`packages/server/src/middlewares/userData.ts:130`) on **every** stream / catalog / search request. Inside it, the precache selector validation block was:

```ts
// validate precache selector
if (config.precacheSelector) {
  try {
    await StreamSelector.testSelect(config.precacheSelector);
  } catch (error) {
    throw new Error(`Invalid precache selector: ${error}`);
  }
}
```

The migration in the same file (and mirrored in `packages/frontend/src/context/userData.tsx`) auto-populates `config.precacheSelector` with `DEFAULT_PRECACHE_SELECTOR` for older configs, even ones with `precacheNextEpisode: false`. So users who have the "Pre-cache Next Episode" toggle **off** still end up with a non-empty `precacheSelector` field, and every request triggers `testSelect` → full SEL eval → log line:

```
SEL | Expression evaluated in 0ms: "count(cached(streams)) == 0 ? uncached(streams) : []"
```

That's exactly what the reporter saw. The actual `precacheNextEpisode()` runtime path in `packages/core/src/main/resources.ts:619` is correctly gated on `ctx.userData.precacheNextEpisode`, so the feature itself doesn't run — only the dormant validation does.

## Fix

One-line gate:

```ts
if (config.precacheSelector && config.precacheNextEpisode) {
```

If the user later flips the toggle on, `validateConfig` re-runs at save time (and on the next request) and will catch any invalid expression before it's used at runtime. Other validated expression fields (`excludedStreamExpressions`, `requiredStreamExpressions`, etc.) don't have a corresponding feature toggle, so they're unaffected.

## Verification

- `pnpm -F core run build` — clean compile.
- Manually traced the eval path: `precacheNextEpisode()` at `packages/core/src/main/resources.ts:619` already gates the runtime path on `ctx.userData.precacheNextEpisode`. No other call site to `precacheSelector` runs in the request hot path; the only caller was `validateConfig`.
- The migration logic at `config.ts:823-834` and `userData.tsx:175-186` is unchanged — users still get the default selector populated, it just no longer gets validated until they enable the feature.

## Risks

- A user could now save a config with an invalid `precacheSelector` while the feature is disabled, then later enable it. The next save / request after enabling will re-run `validateConfig` with the gate satisfied, and the existing throw path (`Invalid precache selector: ...`) will fire — same behavior they'd get if they typed the bad selector after enabling. So the failure mode is preserved, just deferred to when it matters.
- No schema changes, no migration changes, no behavior change for users with the feature on. Build passes.

## Release Notes

* **Bug Fixes**
  * Improved logging performance by reducing unnecessary validation messages when the "Pre-cache Next Episode" feature is disabled.

### DIFF
--- a/packages/core/src/utils/config.ts
+++ b/packages/core/src/utils/config.ts
@@ -511,8 +511,11 @@ export async function validateConfig(
     }
   }
 
-  // validate precache selector
-  if (config.precacheSelector) {
+  // validate precache selector (only when the feature is enabled — the
+  // selector is otherwise dormant, and evaluating it on every request just
+  // pollutes logs with `SEL Expression evaluated...` for users who have the
+  // "Pre-cache Next Episode" toggle off; closes #913)
+  if (config.precacheSelector && config.precacheNextEpisode) {
     try {
       await StreamSelector.testSelect(config.precacheSelector);
     } catch (error) {


Viren070:
Per https://github.com/Viren070/AIOStreams/issues/913#issuecomment-4490241352, I don't really consider this something that needs to be fixed, and if we were going to make this change, other parts of validateConfig should also be updated to match this behaviour e.g. updating groups to not be validated if they are not enabled.

PR #928, opened 2026-05-06
fix(api): stop leaking passwords in URL on user login (#926)

### LINKED ISSUE #926: bug: Plaintext user passwords are viewable on HTTP logs
### Deployment Method

VPS

### Addon Version

v2.29.2

### Bug Area

Configuring

### Device/Browser/OS/Stremio Version

_No response_

### Bug Description / Steps to Reproduce

User logins are conducted over HTTP GET and the plain text password is viewable on HTTP log. 

Creating a user is done over POST and no plaintext password is shown there. 

User creation (GOOD):
`[05/May/2026:21:03:22 -0400] "POST /api/v1/user HTTP/2.0" 201 270 "-" "-" 67962 "aiostreams@docker" `

User login (NOT GREAT):
`[05/May/2026:21:04:11 -0400] "GET /api/v1/user?uuid=c95c8a2e-d163-4162-9ebf-8abbf01e9560&password=ThePasswordText&raw=true HTTP/2.0"`

### Expected Behavior

User password should be viewable as plaintext on log files.

### Configuration Export

Conduct login over HTTP POST with no plain text password. 

### Screenshots

_No response_

### Debugging Checklist

- [x] If applicable, I have included server logs 
- [x] If applicable, I have included MediaFlow logs

### Issue Checklist

- [x] I have written a short but informative title that clearly describes the issue.
- [x] I have given clear and descriptive steps to reproduce the issue.
- [x] I have checked open and closed issues and confirmed that this is not a duplicate of another issue.
- [x] I have filled out all of the requested information adequately. 
- [x] I am using the [latest version](https://github.com/Viren070/AIOStreams/releases/latest).

Sanjays2402:
Closes #926.

## Diagnosis

The login flow uses `GET /api/v1/user?uuid=...&password=...`. Because the password rides in the URL, it ends up in every place HTTP request URLs land:

- reverse-proxy access logs (nginx / caddy / cloudflare in front of Docker — exactly where the reporter caught it)
- browser address bar + history
- HTTP `Referer` headers when the page navigates outbound
- any intermediate CDN / proxy log on the request path

Reporter's nginx log:

```
GET /api/v1/user?uuid=c95c8a2e-...&password=ThePasswordText&raw=true HTTP/2.0
```

The other mutating endpoints (`POST /user`, `PUT /user`, `DELETE /user`, `POST /user/password`, `POST /user/verify`) all carry credentials in the body and don't have this problem. Only the read path was broken.

## Fix

Three coordinated changes:

1. **Server** (`packages/server/src/routes/api/user.ts`) — extract the existing `GET /` handler body into a shared `handleGetUserDetails` helper, then add `POST /load` that calls the same helper with `req.body.{uuid,password,raw}`. The legacy `GET /` handler is preserved (the API is publicly documented; users may have automation hitting it) but now logs a deprecation warning on every call:

   ```
   Deprecated: GET /api/v1/user exposes the password in the request URL
   (visible in HTTP access logs, browser history, Referer headers).
   Use POST /api/v1/user/load instead. See https://github.com/Viren070/AIOStreams/issues/926
   ```

2. **Frontend** (`packages/frontend/src/lib/api.ts`) — switch `loadUserConfig` and `loadRawUserConfig` from `GET /user?uuid=...&password=...` to `POST /user/load` with a JSON body. Same return shape, no callers had to change.

3. **Alias middleware** (`packages/server/src/middlewares/alias.ts`) — extend the POST branch so `req.body.uuid` aliases resolve on POST endpoints. This is needed for `POST /load` (otherwise users with `ALIASED_CONFIGURATIONS` set would lose alias support compared to the GET form) and incidentally fixes the silent gap on `POST /verify` and `POST /password`. `POST /api/v1/user` (create) has no `uuid` in the body, so the lookup no-ops there.

4. **Docs** (`packages/docs/content/docs/apis/user.mdx`) — lead with `POST /api/v1/user/load` as the recommended form, keep the `GET` form documented as **deprecated** with a callout linking to #926.

## Verification

- `pnpm -F core run build`, `pnpm -F server run build`, `pnpm -F frontend run build` — all clean.
- Traced both call paths to confirm the shared helper produces the same response shape on both methods (same `userData` + `encryptedPassword` payload, same error mapping for `MISSING_REQUIRED_FIELDS`, `ENCRYPTION_ERROR`, `INTERNAL_SERVER_ERROR`).
- `raw` parameter accepts string `'true'` (legacy GET, query strings are always strings) **or** boolean `true` (new POST, JSON), so frontend can send `raw: true` directly without the `'true'` quoting.
- Frontend rate limiter `userApiRateLimiter` and the alias middleware are still applied at the router level, so they cover both methods.
- No DB schema changes, no migration changes, no environment variable changes.

## Risks

- **Backward compat**: the deprecated GET endpoint still works exactly as before. Existing automation, manifest URL flows, and any third-party clients keep functioning. The deprecation warning is server-side only (log level `warn`), it doesn't change responses.
- **Alias resolution on other POST endpoints**: previously `POST /verify` and `POST /password` accepted `req.body.uuid` directly without alias resolution. Now they do alias resolution, which is consistency-positive (the docs already claim "Wherever a UUID is accepted, you may supply a configured alias"). If anyone had a UUID that happened to also be a configured alias key, the resolver short-circuits on UUID format (`uuidRegex.test(value)`) so the existing behavior is preserved for actual UUIDs.
- **No password change is required from users** — the password value itself is unchanged, only its transport.

## Follow-ups (not in this PR)

- Consider eventually removing `GET /api/v1/user` after a deprecation window. Not in scope here.
- Audit any logs that might persist `req.query` or full request URLs server-side (out of scope; the reported leak is in the reverse proxy, which this PR fixes).

## Release Notes

* **New Features**
  * Added POST /api/v1/user/load endpoint to retrieve full user configuration and encrypted password.
  * Added POST endpoint for user account creation.

* **Documentation**
  * Updated API documentation with new endpoint specifications and migration guidance.
  * Marked GET /api/v1/user as deprecated whilst maintaining full backward compatibility for existing integrations.

### DIFF
--- a/packages/docs/content/docs/apis/user.mdx
+++ b/packages/docs/content/docs/apis/user.mdx
@@ -56,16 +56,25 @@ curl -X HEAD "https://your-instance.example.com/api/v1/user?uuid=xxxxxxxx-xxxx-x
 
 ## Get User
 
-`GET /api/v1/user`
+`POST /api/v1/user/load`
 
-Returns the full user configuration and an encrypted copy of the password for use in subsequent requests.
+Returns the full user configuration and an encrypted copy of the password for use in subsequent requests. Prefer this form over `GET /api/v1/user` — the password is sent in the JSON body so it never appears in URLs, HTTP access logs, browser history, or `Referer` headers.
 
-### Query Parameters
+### Request Body
 
-| Parameter  | Type   | Required | Description                 |
-| ---------- | ------ | -------- | --------------------------- |
-| `uuid`     | string | ✅       | UUID (or alias) of the user |
-| `password` | string | ✅       | User's plaintext password   |
+```json
+{
+  "uuid": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
+  "password": "mypassword",
+  "raw": false
+}
+```
+
+| Field      | Type    | Required | Description                                                                                |
+| ---------- | ------- | -------- | ------------------------------------------------------------------------------------------ |
+| `uuid`     | string  | ✅       | UUID (or alias) of the user                                                                |
+| `password` | string  | ✅       | User's plaintext password                                                                  |
+| `raw`      | boolean | ❌       | When `true`, returns the raw stored config without server-side hydration. Defaults to `false`. |
 
 ### Response `200`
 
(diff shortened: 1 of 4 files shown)


Viren070:
Hey, thanks for the PR but I went ahead and reworked the User API to use HTTP basic authentication instead.

PR #940, opened 2026-05-10
fix(builtins/torrentgalaxy): use different domain by default

webstreamr:
.space does not work anymore, the domain's SSL is misconfigured it seems

## Summary by CodeRabbit

* **Chores**
  * Updated the default Torrent Galaxy addon endpoint to reflect the current service URL.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Viren070/AIOStreams/pull/940)


jakubschenk:
+1 Can confirm that it no longer works and that .one works

PR #952, opened 2026-05-14
fix(core): redact apikey in nab keys + http recursion logs (follow-up #947)

slawrensen:
Follow-up to #947. That PR landed the Newznab error/debug log redaction but left a few of the review comments unaddressed. This picks those up, plus one related surface I noticed while looking at the same paths again.

Changes (3, smallest possible):

- `packages/core/src/utils/http.ts` — `makeUrlLogSafe` docstring updated. The function already strips `apikey=` (line 46) but the comment only mentioned `password`. Doc now matches behavior.

- `packages/core/src/utils/http.ts` — `PossibleRecursiveRequestError` warn and throw embedded the raw `urlObj.toString()`, while the sibling DEBUG log on the same path already runs through `makeUrlLogSafe`. Wrapped both so recursion errors don't leak apikey for any caller (not just nab).

- `packages/core/src/builtins/base/nab/api.ts` — `cacheKey` and `lockKey` embedded raw `this.apiKey`. Both keys flow into `logger.*` calls inside `DistributedLock` and `searchWithBackgroundRefresh`. Hashed with the existing `getSimpleTextHash` helper so credential partitioning is kept but the raw key never reaches a log line.

Note: the cache/lock key change is a one-time cache-bust on deploy (existing entries won't match the new key shape). No data loss, just a cold cache for nab search/caps.

Outbound request behavior is unchanged — the apikey is still sent in plain to the upstream Newznab endpoint via `searchParams.set('apikey', ...)`.

No new deps, no public API change, no tests added (matching #947). Happy to add tests if you want them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced security by hashing API keys used in internal caches and distributed locks to avoid exposing sensitive credentials.
  * Improved log safety by masking password and API key parameters (including URL userinfo) to prevent accidental disclosure in logs and error messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Viren070/AIOStreams/pull/952)

### DIFF
--- a/packages/core/src/builtins/base/nab/api.ts
+++ b/packages/core/src/builtins/base/nab/api.ts
@@ -4,6 +4,7 @@ import {
   DistributedLock,
   Env,
   formatZodError,
+  getSimpleTextHash,
   getTimeTakenSincePoint,
   createLogger,
   makeRequest,
@@ -369,7 +370,7 @@ export class BaseNabApi<N extends 'torznab' | 'newznab'> {
     searchFunction: string = 'search',
     params: Record<string, string | number | boolean> = {}
   ): Promise<SearchResponse<N>> {
-    const cacheKey = `${this.baseUrl}${this.apiPath}?t=${searchFunction}&${JSON.stringify(params)}&apikey=${this.apiKey}&${JSON.stringify(this.params)}`;
+    const cacheKey = `${this.baseUrl}${this.apiPath}?t=${searchFunction}&${JSON.stringify(params)}&apikey=${this.apiKey ? getSimpleTextHash(this.apiKey) : ''}&${JSON.stringify(this.params)}`;
 
     return searchWithBackgroundRefresh({
       searchCache: this.searchCache as Cache<string, SearchResponse<N>>,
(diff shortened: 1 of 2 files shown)


slawrensen:
Pushed an amend addressing two more leak surfaces on the same code path:

- `makeUrlLogSafe` now parses the URL and redacts userinfo (`https://user:apikey@host`) and any `password`/`apikey` param in the **fragment** as well as the query, case-insensitive. Legacy regex stripping is kept as a fallback for non-URL strings (already-truncated log fragments etc.). Also tightened the `password=` regex to `/gi` so it matches the `apikey=` line.
- `makeRequest` debug log now runs `options.forceProxy` through `makeUrlLogSafe` before logging, so a proxy URL with userinfo doesn't leak.

Same scope (#947 follow-up, leak redaction), no behavior change for outbound traffic, no new deps.

Out of scope but worth flagging for a separate PR: `startup.ts:140` logs `Env.ADDON_PROXY.join(', ')` raw at boot, same bug class.

PR #974, opened 2026-05-23
fix: pass streams parameter to toFormatterContext in ChillLink route

curiounomad:
The toFormatterContext() call was missing the streams parameter, preventing calculation of maxSeScore and maxRegexScore. This caused normalized score variables (nSeScore, nRegexScore) to return null in ChillLink.

This matches the implementation in the Stremio route which correctly passes response.data.streams.


Viren070:
Thanks!

PR #978, opened 2026-05-24
fix(presets/stremthru-torz): bring back overriden stream parser

webstreamr:
was removed by mistake in 1a9892b

fixes things like private torrent and separate audio/subtitle language detection

untested YOLO 🙈

## Release Notes

* **Refactor**
  * Restructured internal stream parsing architecture with a dedicated parser implementation to enhance performance and code maintainability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Viren070/AIOStreams/pull/978?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/presets/stremthruTorz.ts
+++ b/packages/core/src/presets/stremthruTorz.ts
@@ -10,9 +10,21 @@ import { baseOptions, Preset } from './preset.js';
 import { constants, ServiceId } from '../utils/index.js';
 import { config as appConfig } from '../config/index.js';
 import { StreamParser } from '../parser/index.js';
-import { StremThruPreset } from './stremthru.js';
+import { StremThruPreset, StremThruStreamParser } from './stremthru.js';
+
+class StremthruTorzStreamParser extends StremThruStreamParser {
+  protected override applyUrlModifications(
+    url: string | undefined
+  ): string | undefined {
+    return super.applyUrlModifications(url);
+  }
+}
 
 export class StremthruTorzPreset extends StremThruPreset {
+  static override getParser(): typeof StreamParser {
+    return StremthruTorzStreamParser;
+  }
+
   static override get METADATA() {
     const supportedResources = [constants.STREAM_RESOURCE];
 


EZE404:
merged subs languages within audio languages still happening for stremthru torz using v2.30.2:

imagen

formatter sentence used for languages emojis:
`{stream.languageEmojis::exists["📢 [{stream.languageEmojis::join('][')::replace('💃🏻', '🇲🇽')}]"||""]}`

real video content: only english audio with a lot of subtitles languages embedded

webstreamr:
> merged subs languages within audio languages still happening for stremthru torz using v2.30.2:
> 
> imagen
> formatter sentence used for languages emojis: `{stream.languageEmojis::exists["📢 [{stream.languageEmojis::join('][')::replace('💃🏻', '🇲🇽')}]"||""]}`
> 
> real video content: only english audio with a lot of subtitles languages embedded

It looks like stremthru only gets subs for that one, see

Screenshot_20260527-073819_1

The reason is that the audio is tagged as "Original" and seemingly skipped. If stremthru would interpret this as English in this case it would be perfect, FYI @MunifTanjim 

Further down the line in aios, it seems that it throws away the subs only info and merges it as general language info. Not sure if we should behave like stremthru and only show subs instead FYI @Viren070 

But this is definitely a bit of an edge case.

I think I'll look into adapting the aios parser, so that it supports audio/subs only info too. But I'm afraid that we can't pass through "Original" easily and stremthru might also not have the meta data to interpret what Original means.

EZE404:
> The reason is that the audio is tagged as "Original" and seemingly skipped. If stremthru would interpret this as English in this case it would be perfect, FYI @MunifTanjim
> 
> Further down the line in aios, it seems that it throws away the subs only info and merges it as general language info. Not sure if we should behave like stremthru and only show subs instead FYI @Viren070
> 
> But this is definitely a bit of an edge case.
> 
> I think I'll look into adapting the aios parser, so that it supports audio/subs only info too. But I'm afraid that we can't pass through "Original" easily and stremthru might also not have the meta data to interpret what Original means.

In that case, wouldn't TMDB integration within AIOs be useful for identifying the original audio language?

webstreamr:
> In that case, wouldn't TMDB integration within AIOs be useful for identifying the original audio language?

Munif (stremthru developer) responded via Discord, explaining that ffmpeg is not getting any language info at all there. in that case we can't do anything I'm afraid. We also can't assume that it is the default language, but in most cases it would be I suppose. But I'll adapt aios to at least not shuffle audio and subs together in such rare (I think?) cases.

PR #1016, opened 2026-06-13
fix(marketplace): remove deprecated usa tv and debridio tv addons

i4mth3d4ng3r:
removed usa tv and debridio tv due to deprecation

## Summary by CodeRabbit

* **Chores**
  * Removed Debridio TV preset along with its configuration options and environment variables.
  * Removed USA TV preset along with its configuration options and environment variables.

* **Documentation**
  * Updated environment variables reference to reflect the removal of Debridio TV and USA TV configuration options.

### DIFF
--- a/packages/core/src/builtins/newznab/addon.ts
+++ b/packages/core/src/builtins/newznab/addon.ts
@@ -192,6 +192,7 @@ export class NewznabAddon extends BaseNabAddon<NewznabAddonConfig, NewznabApi> {
         age: age,
         title: result.title,
         indexer:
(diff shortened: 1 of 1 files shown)


Viren070 (changes requested):
it is better to use the `DISABLED` and set `removed` within to true, this way we can gracefully handle configs which still have this addon installed, like so: 

https://github.com/Viren070/AIOStreams/blob/4894cde55202f648d87ba8e006a634cddc5c3f11/packages/core/src/presets/torbox.ts#L174-L178

PR #1020, opened 2026-06-16
feat(subtitles): proxy subtitle URLs through the configured proxy

IbbyLabs:
## Release Notes

* **New Features**
  * Added support for proxying subtitle URLs. You can now enable subtitle proxying in the proxy settings menu and choose which subtitle URLs are proxied globally or per addon instance, working alongside existing stream proxying.
* **Configuration**
  * Updated proxy configuration descriptions to clarify subtitle URL proxying and how the “subtitle” service selection affects behaviour.

### DIFF
--- a/packages/core/src/config/schema/proxy.ts
+++ b/packages/core/src/config/schema/proxy.ts
@@ -133,7 +133,7 @@ export const proxySchema = {
       default: null as string[] | null,
       label: 'Default proxied services',
       description:
-        'List of serviceIds to proxy by default. JSON array of strings.',
+        'List of serviceIds to proxy by default. Use `subtitle` to proxy subtitle URLs. JSON array of strings.',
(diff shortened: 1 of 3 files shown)


Viren070 (changes requested):
It would be better to add an option for it - not everyone would want to proxy subtitles.

IbbyLabs:
> It would be better to add an option for it - not everyone would want to proxy subtitles.

Valid. It’s quite niche tbh and only really affects self hosters. I’ll look at it again and see if it’s worth doing or dropping. Cheers virenio

IbbyLabs:
@Viren070 is this better boss

IbbyLabs:
Folded into #1029, which now covers subtitle-URL proxying (both auto-proxying internal-host subtitles and the opt-in `subtitle` proxied-service). Closing in favour of that single proxy PR.

PR #1022, opened 2026-06-16
fix anime split-cour stream matching

devils-shadow:
## Summary

Fix anime stream matching for split-cour / sequel entries where the Stremio request ID uses one external season number, but anime databases and release titles identify the requested content as a later logical season or cour.

The motivating case is *Bungo Stray Dogs*: Stremio metadata can expose three seasons, while anime trackers commonly list the same later episodes as separate 4th and 5th season entries. Requests such as `tt5679720:3:*` can therefore need streams titled `Bungo Stray Dogs 4th Season` / `Bungo Stray Dogs 5th Season`, not generic season 3 releases.

## Problem

AIOStreams currently treats `season` / `episode` as a single coordinate system. That works for most TV metadata, but it breaks when anime metadata has two valid coordinate systems:

- External request coordinates from Stremio / IMDb / TMDB / TVDB, for example `tt5679720:3:14`.
- Logical anime entry coordinates from anime mappings and release titles, for example `Bungo Stray Dogs 4th Season`, episode 1.

When those are collapsed into one value, several downstream stages can become incorrect:

- Search queries may ask indexers for the external season instead of the anime sequel title or logical season.
- Season/episode filtering may discard otherwise correct releases because parsed files contain logical season numbers.
- Broad query fanout can over-query Torznab/Newznab endpoints when the first query form does not match.
- Repeated 429 responses can continue to be retried immediately instead of backing off.

## Approach

This PR preserves both coordinate systems instead of replacing one with the other:

- `logicalSeason` / `logicalEpisode`: anime-entry numbering inferred from anime entry titles and episode offsets.
- `externalSeason` / `externalEpisode`: original request coordinates from the parsed Stremio ID.
- Stream selection and debrid metadata can use logical coordinates where that is necessary, while formatter/expression contexts expose logical and external values explicitly.

For ambiguous anime seasons, the search metadata now keeps sequel-specific anime titles and marks the request for query search. Nab-based addons then use bounded query waves:

1. Base title + logical `SxxExx` / `Sxx`.
2. Anime entry title + logical episode forms.

The second wave only runs if the first wave returns no results. This keeps the fix targeted without spraying every possible title/episode variant at indexers.

If query search is required but the selected Nab search function does not support `q`, the request now fails fast instead of issuing a broad unqualified search.

## Nab rate-limit handling

Torznab/Newznab APIs now track endpoint-level 429 cooldowns. During cooldown:

- cached search results can still be returned without refresh;
- uncached searches fail fast with a rate-limit error;
- query batches stop treating 429s as an invitation to continue fanout.

This is intentionally generic and applies to Nab endpoints broadly, not to one indexer-specific workaround.

## Filtering and expressions

Season/episode filtering now prefers logical anime coordinates while still allowing an exact external season/episode tuple when a release is parsed that way.

Formatter and expression contexts also expose the explicit logical/external values, so advanced users can reason about either coordinate system without overloading `season` and `episode`.

## Validation

Added deterministic unit coverage for the pure anime search helpers:

- extracts logical seasons from sequel titles such as `4th Season`, `Season 2`, and roman numeral forms;
- converts external episode offsets into entry-relative logical episodes;
- keeps sequel-specific titles while deriving base-title candidates for bounded query search.
- verifies Bungo-style split-cour query waves such as `Bungo Stray Dogs S04E01` and `Bungo Stray Dogs 4th Season 01`.

Validated locally with:

```bash
pnpm -F core test
pnpm -F core run build
```

No tests call live indexers or content endpoints.

## Summary by CodeRabbit

* **New Features**
  * Improved anime season/episode handling by distinguishing logical vs external numbering for more accurate ambiguous-release searches.
  * Added logical/external season/episode metadata support across parsing, formatter and expression contexts.
  * Enhanced debrid query generation with configurable season/episode query variants, deduped/cleaned titles, and bounded “max titles” behavior.
  * Upgraded NAB searching with optional prefer-query-search routing, wave-based query progression, and explicit rate-limit handling with retry timing.

* **Bug Fixes**
  * Preserved valid Stremio coordinate value `0` and improved season/episode matching when external season data is present.

* **Tests**
  * Added/extended Vitest coverage for anime search metadata helpers and Stremio coordinate parsing.

### DIFF
--- a/packages/core/src/builtins/base/debrid.ts
+++ b/packages/core/src/builtins/base/debrid.ts
@@ -6,7 +6,12 @@
   Stream,
 } from '../../db/schemas.js';
 import { z, ZodError } from 'zod';
-import { IdParser, IdType, ParsedId } from '../../utils/id-parser.js';
+import {
+  IdParser,
+  IdType,
+  ParsedId,
+  parseStremioCoordinate,
+} from '../../utils/id-parser.js';
 import {
   AnimeDatabase,
   BuiltinServiceId,
@@ -205,8 +210,8 @@ export abstract class BaseDebridAddon<T extends BaseDebridConfig> {
       this._searchMetadataPromise = Promise.resolve({
         primaryTitle: undefined,
         titles: [],
-        season: parsedId.season ? Number(parsedId.season) : undefined,
-        episode: parsedId.episode ? Number(parsedId.episode) : undefined,
+        season: parseStremioCoordinate(parsedId.season),
+        episode: parseStremioCoordinate(parsedId.episode),
       });
     }
 
@@ -549,8 +554,12 @@ export abstract class BaseDebridAddon<T extends BaseDebridConfig> {
     }
 
     const titlePlaceholder = '<___title___>';
-    const querySeason = metadata.logicalSeason ?? parsedId.season;
-    const queryEpisode = metadata.logicalEpisode ?? parsedId.episode;
+    const querySeason = parseStremioCoordinate(
+      metadata.logicalSeason ?? parsedId.season
+    );
+    const queryEpisode = parseStremioCoordinate(
+      metadata.logicalEpisode ?? parsedId.episode
+    );
     const externalSeason = metadata.externalSeason;
     const externalEpisode = metadata.externalEpisode;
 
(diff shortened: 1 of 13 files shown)


devils-shadow:
forgot to add, I've been using this fix on my instance for the past week without any obvious issues, tested mainly with anime-type media.

PR #1025, opened 2026-06-18
feat: include finished group results on early exit

dan-dr:
## Summary
- add a parallel group option to include already-finished group queries when an early-exit condition fails
- persist/validate the new group setting and expose it in the addon fetching UI
- document the parallel finished-results behavior
- convert a few StreamContext/ExpressionContext imports to type-only imports while reviewing testability

## Suggested follow-up tests
- Core: add a focused parallel-groups test with three mocked addon timings to assert default early exit excludes settled fallback results and the new flag includes only already-settled results.
- Frontend: add a small component test for the switch only if/when the frontend has an existing React component test harness; currently typecheck/build coverage is the lightweight fit.

## Verification
- pnpm -F core run build
- pnpm -F frontend typecheck
- pnpm -F frontend run build
- git diff --check

Note: commands ran under Node v22.21.0, so pnpm printed the repo's Node >=24 engine warning.

## Release Notes

**New Features**
* Added "Always add results from finished queries" option for parallel group fetching. When enabled, results from completed queries are included in the output even if a group condition fails partway through execution, rather than discarding pending results entirely.

**Documentation**
* Updated group fetching documentation to clarify behaviour when group conditions evaluate to false and explain the impact of the new finished results option.

### DIFF
--- a/packages/core/src/db/schemas.ts
+++ b/packages/core/src/db/schemas.ts
@@ -660,6 +660,7 @@ export const UserDataSchema = z.object({
         )
         .optional(),
       behaviour: z.enum(['sequential', 'parallel']).optional(),
+      includeFinishedResultsOnEarlyExit: z.boolean().optional(),
     })
     .optional(),
   sortCriteria: z.object({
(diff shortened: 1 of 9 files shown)


dan-dr:
Hello mr viren @Viren070

dan-dr:
I'm telling you, this is a killer feature.

PR #1034, opened 2026-06-21
add "Delete All" button to clear the entire library

Djangowaming:
Adds a destructive "Delete All" button to the Usenet library toolbar that removes every entry from the library with a single confirmation dialog, complementing the existing per-entry and multi-select delete workflows.

### Changes
 packages/core/src/db/repositories/usenet-library.ts 
- Added  UsenetLibraryRepository.clear()  — a static method that issues  DELETE FROM usenet_library  (no filter) and emits the  change  event so the dashboard SSE stream pushes the update live.
 packages/server/src/routes/api/dashboard-usenet.ts 
- Added  DELETE /dashboard/usenet/library  route (registered before the existing  DELETE /library/:hash  parametric route so Express matches the fixed path first) that delegates to  UsenetLibraryRepository.clear() .
 packages/frontend/src/app/dashboard/usenet/queries.ts 
- Added  useDeleteAllLibraryEntries()  mutation — a no-argument TanStack Query mutation hitting the new batch endpoint, with automatic cache invalidation on success.
 packages/frontend/src/app/dashboard/usenet/library-page.tsx 
- Placed a "Delete all" icon button (trash icon,  alert-subtle  intent) in the toolbar next to the "Select" button, visible outside of select mode.
- Tapping it opens a  ConfirmationDialog  with the message "This will remove every entry from the library, including active imports. This action cannot be undone.".
- On confirm, calls the batch delete mutation with proper success/error toast feedback and resets any active selection state.

Usenet library dashboard icon update for frontend UI:
image

Confirmation prompt:
image

### DIFF
--- a/packages/core/src/db/repositories/usenet-library.ts
+++ b/packages/core/src/db/repositories/usenet-library.ts
@@ -387,6 +387,12 @@ export class UsenetLibraryRepository {
     usenetLibraryBus.emit('change');
   }
 
+  /** Remove every entry from the library. */
+  static async clear(): Promise<void> {
+    await getDb().exec(sql`DELETE FROM usenet_library`);
+    usenetLibraryBus.emit('change');
+  }
+
   /** The streamable file list for an entry. */
   static async getFiles(nzbHash: string): Promise<UsenetLibraryFile[]> {
     const entry = await this.get(nzbHash);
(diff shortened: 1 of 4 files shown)


Djangowaming:
@Viren070 any issues you see with this PR? Happy to answer any questions.

Viren070 (approved):

PR #1035, opened 2026-06-22
Fix misleading latency in Usenet provider "Test connection"

Djangowaming:
The "Test connection" button on the dashboard Usenet providers page was reporting a  latencyMs  value that included the entire connection lifecycle (TCP handshake + TLS negotiation + server greeting parsing + AUTHINFO USER/PASS exchange + DATE command), inflating the number to ~6× the actual network round-trip time.
For example, a provider with 20ms network ping would show ~120ms. Naturally, users interpreted this as the server being slow.

ignore if this isn’t the intention of the test connection button.

### The fix
Moved the  const start = Date.now()  timer to after  NntpConnection.connect()  completes, so  latencyMs  now measures only the DATE command round-trip — one network round-trip on an already-established connection.
The test still validates the full path (TCP, TLS, greeting, auth all run and throw fast on failure), but the displayed latency now reflects actual server responsiveness rather than connection setup overhead.

### Changes
 packages/core/src/usenet/integration/dashboard/providers.ts  — moved the timer start inside the  try  block, after  connect()  succeeds, and updated the JSDoc to document what  latencyMs  represents.

### DIFF
--- a/packages/core/src/usenet/integration/dashboard/providers.ts
+++ b/packages/core/src/usenet/integration/dashboard/providers.ts
@@ -91,7 +91,11 @@ export async function saveUsenetProviders(
   await settingsStore.set('usenet.providers', merged, username);
 }
 
-/** Test a single provider connection (dial + auth + DATE health probe). */
+/** Test a single provider connection (dial + auth + DATE health probe).
+ *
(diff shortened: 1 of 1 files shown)


Djangowaming:
@Viren070 any issues you see with this PR? Happy to answer any questions.

PR #1041, opened 2026-06-26
fix(prowlarr): add yearless movie queries

needforseed1:
## Summary
- include title-only Prowlarr queries for movie searches in addition to title + year
- deduplicate generated queries before dispatching searches

## Why
Some movie metadata can have an incorrect or future year compared to indexed releases. Searching only with title + year can return no usable Prowlarr results even when title-only results exist.

## Testing
- pnpm -F core run test
- pnpm -F core run build
- pnpm exec prettier --check packages/core/src/builtins/prowlarr/addon.ts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search handling for movies by trying an additional match variation, which can help find more results.
  * Removed duplicate search attempts and skips searching when no valid queries are available.


webstreamr:
This could be a general znab feature to configure /opt-in per add-on instead of being hard-wired for prowlarr only?

PR #1059, opened 2026-07-02
Detect missing segments mid stream

Djangowaming:
Problem:
There is a way for usenet streams to get past the health check even though it has missing segments.

solution:
Add a listener that will watch for missing segment errors and mark the link as bad so failover can kick in the next time that stream is selected

note:
For whatever reason this allows for mid stream failover. I suspect it will only work if the missing segments are caught at the beginning on the stream rather than at the middle or end.


Viren070:
implemented similar logic in  https://github.com/Viren070/AIOStreams/commit/678b42e24565495f191a2314be04e56b3105465c , thanks!

PR #1063, opened 2026-07-03
feat(screener): release keys and NDJSON interchange (1/6)

IbbyLabs:
Part 1 of the Screener feature (Warden for AIOStreams), split out from #1056 for review.

This is the release identity everything else is built on, plus the portable list format:

- `btih:<infohash>` for torrents, a nzbdavex/Warden-compatible `wd1:<fingerprint>` for usenet. The fingerprint is credential-free (`sha256(size|poster|day)`, first 16 bytes), so a shared list carries only digests, no titles/URLs/account info.
- NDJSON import/export that reads and writes both aiostreams-native and Warden lists, which is what lets the two ecosystems share one pool.

Pure functions, no pipeline wiring. Unit tests cover fingerprint/key parity with Warden and NDJSON round-trips in both dialects.

Contains only its own files, so it won't build standalone. Merge order is 1 → 6; see #1056 for the full feature.
---
**Screener, split from #1056 into 6 parts to review in passes. Merge in order:**
1. #1063 — release keys + NDJSON interchange
2. #1064 — verdict store + migrations
3. #1065 — evaluator + config
4. #1066 — pipeline filter
5. #1067 — API + remote + GitHub sync
6. #1068 — dashboard

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -2,6 +2,7 @@ export * from './utils/index.js';
 export * from './logging/ring-buffer.js';
 export * from './config/index.js';
 export * from './db/index.js';
+export * from './screener/index.js';
 export * from './analytics/index.js';
 export * from './analytics/repository.js';
 export * from './tasks/index.js';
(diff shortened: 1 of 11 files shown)


IbbyLabs:
Folding this back into #1056 as a single branch. Thanks!

PR #1065, opened 2026-07-03
feat(screener): trust/quorum/backbone evaluator and config (3/6)

IbbyLabs:
Part 3 of the Screener feature, split from #1056.

The decision logic plus config. Turns stored verdicts into a filter decision:

- `full` / `corroborate` / `observe` trust levels and a corroboration quorum.
- Optional backbone scoping: a verdict from a provider backbone that isn't yours can't hide a release that's still fine on yours. Root-domain collapsing is ported from Warden's `RootDomain`.
- The per-user config schema, and the operator backbone env vars (documented).

Still headless, no wiring into the pipeline yet. Evaluator rules are unit-tested.

Depends on parts 1-2. Merge order 1 → 6; umbrella is #1056.
---
**Screener, split from #1056 into 6 parts to review in passes. Merge in order:**
1. #1063 — release keys + NDJSON interchange
2. #1064 — verdict store + migrations
3. #1065 — evaluator + config
4. #1066 — pipeline filter
5. #1067 — API + remote + GitHub sync
6. #1068 — dashboard

PR #1064, opened 2026-07-03
feat(screener): verdict store and migrations (2/6)

IbbyLabs:
Part 2 of the Screener feature, split from #1056.

Persistence for verdicts: the repository and migrations `0012`/`0013`, on the same SQLite+Postgres driver abstraction as the rest of the DB layer. Stores per-source verdicts keyed by the release identity from part 1. No filtering yet. SQL-equivalence tests run the repository against both backends.

Depends on part 1. Merge order 1 → 6; umbrella is #1056.
---
**Screener, split from #1056 into 6 parts to review in passes. Merge in order:**
1. #1063 — release keys + NDJSON interchange
2. #1064 — verdict store + migrations
3. #1065 — evaluator + config
4. #1066 — pipeline filter
5. #1067 — API + remote + GitHub sync
6. #1068 — dashboard

PR #1066, opened 2026-07-03
feat(screener): filter flagged releases in the stream pipeline (4/6)

IbbyLabs:
Part 4 of the Screener feature, split from #1056.

Where Screener goes live in the stream pipeline:

- Drops flagged releases before dedup/failover, recorded as a `screener` filter reason so it shows in per-request stats, via a `screenFlagged` step in the filterer.
- Skips a sibling NZB of a release already proven dead at playback resolve, under the viewer's own quorum/backbone.
- Auto-marks a usenet release `dead` when the native engine confirms it's gone from every provider.
- Builtins precompute the `wd1` key from indexer size/poster/date so the filter never re-derives it.

Fails open throughout: any store error passes streams through, and it never leaves a viewer with nothing (if every stream is flagged, all are shown).

Depends on parts 1-3. Merge order 1 → 6; umbrella is #1056.
---
**Screener, split from #1056 into 6 parts to review in passes. Merge in order:**
1. #1063 — release keys + NDJSON interchange
2. #1064 — verdict store + migrations
3. #1065 — evaluator + config
4. #1066 — pipeline filter
5. #1067 — API + remote + GitHub sync
6. #1068 — dashboard

### DIFF
--- a/packages/core/src/builtins/base/debrid.ts
+++ b/packages/core/src/builtins/base/debrid.ts
@@ -36,6 +36,7 @@ import {
   metadataStore,
   fileInfoStore,
   FileInfo,
+  ScreenerOptionsSchema,
 } from '../../debrid/index.js';
 import { processTorrents, processNZBs } from '../utils/debrid.js';
 import { calculateAbsoluteEpisode } from '../utils/general.js';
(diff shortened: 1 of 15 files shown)


IbbyLabs:
Folding this back into #1056 as a single branch. Thanks!

PR #1069, opened 2026-07-03
feat(db): boot older builds on a newer schema when migrations are backward-compatible

IbbyLabs:
AIOStreams migrations are append-only, so once a database is migrated by a newer build, an older build refuses to start (`Database is at migration N but this build only knows up to M`), even when the newer migrations only added things the older code never touches. Rolling back, or trying a newer build and then going back, means manual SQL or wiping the database.

A migration can now declare `backwardCompatible` when it only adds tables/columns/indexes older code ignores. The runner records that per row in `_migrations`, and on boot an older build starts against a newer database only if every newer migration is marked compatible. A rename, drop, or retype of an existing table still blocks the boot.

So rolling back a bad upgrade, or trying a build without committing to it, just works for additive changes and stays protected against the ones that would actually break. Nothing to configure.

### DIFF
--- a/packages/core/src/db/migrations/runner.ts
+++ b/packages/core/src/db/migrations/runner.ts
@@ -16,9 +16,16 @@ async function ensureMigrationsTable(driver: DbDriver): Promise<void> {
     `CREATE TABLE IF NOT EXISTS _migrations (
         id INTEGER PRIMARY KEY,
         name TEXT NOT NULL,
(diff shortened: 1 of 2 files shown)


IbbyLabs:
@Viren070 reworked it, migrations now declare if they're backwards compatible and the runner enforces it, breaking ones still block.

PR #1087, opened 2026-07-12
feat(release-blocklist): share verdicts via a GitHub gist

IbbyLabs:
Sharing a blocklist today means turning on the public export endpoint and giving subscribers a URL on your own instance. That exposes the instance's address to everyone you share with and needs it reachable from the internet, which not everyone wants to do just to publish a list.

This adds an alternative: point the instance at a GitHub token (with `gist` scope) and it publishes its **own** verdicts to a secret gist, then shares the `gist.githubusercontent.com` raw URL. Subscribers get a stable URL that has nothing to do with your instance, and the gist works exactly like every community list people already subscribe to.

How it works:
- Two schema-driven settings under Release Blocklist: `githubToken` (secret) and `githubGistUrl` (auto-filled on first publish; set it yourself to reuse an existing gist).
- `publishBlocklistGist()` creates the gist on first use (`public: false`) and `PATCH`es it in place afterwards, writing the resolved raw URL back to settings.
- Auto-sync piggybacks on the existing `release-blocklist-refresh` task: it only pushes when the local export revision has actually changed, so it's a no-op most ticks and when sharing is off.
- Manual "Publish now" button + copyable URL in the share modal.

Scope is always this instance's own verdicts (`local`), same as the default public export, so you never re-publish other people's lists.

## Summary

* **New Features**
  * Added GitHub Gist sharing for release blocklists.
  * Supports creating a new private gist or updating an existing one.
  * Dashboard now includes a GitHub gist section with copy-to-clipboard and “publish now/update now” actions.
  * The blocklist can now synchronise to the gist automatically after scheduled refreshes.

* **Documentation**
  * Documented `RELEASE_BLOCKLIST_GITHUB_TOKEN` and `RELEASE_BLOCKLIST_GITHUB_GIST_URL` environment variables for configuring gist publishing.


IbbyLabs:
Prebuilt image for testing (multi-arch, amd64 + arm64):

```
ghcr.io/ibbylabs/aiostreams:dev
```

To try it: set a GitHub token with `gist` scope under Settings → Release Blocklist, then hit "Publish now" in the Share modal on the blocklist sources page. It creates a secret gist and shows the raw URL; it also re-publishes on the refresh task when your local verdicts change.

IbbyLabs:
Pushed fixes:
- GitHub requests are now time-boxed (15s) so a hung call can't stall the refresh task.
- On a 404 updating the gist (deleted upstream) it recreates a fresh one instead of failing every sync.
- Whitespace-only tokens are treated as unset in the snapshot and publish route, matching the publisher.

Left `lastSyncedRevision` in memory rather than persisting it: a PATCH is idempotent, so at most one redundant republish happens after a restart, which isn't worth a stored-state migration.

IbbyLabs:
Done in e052cad8292b76b38171b836943de89450ff4b0b

PR #1127, opened 2026-07-22
fix: make request body size limit configurable, raise default to 1mb

funkypenguin:
## Problem

A user reported being unable to save their configuration — the save request failed with:

```
PayloadTooLargeError: request entity too large
```

Their config payload was ~102KB. `express.json()` / `express.urlencoded()` are registered with no options in `packages/server/src/app.ts`, so they inherit Express' built-in **100kb** body limit. Any config with enough addons/groups/regex filters to push the JSON past 100KB becomes unsaveable, and the limit isn't tunable.

## Fix

- New env var `MAX_REQUEST_BODY_SIZE` (default `1mb`), exposed via the bootstrap config (it's needed at middleware-registration time, before the settings store is up).
- Applied to both `express.json()` and `express.urlencoded()`.
- Documented in `.env.sample`.

The 1mb default matches nginx's default `client_max_body_size`, so typical reverse-proxy setups won't need any accompanying proxy change.

## Verification

- `pnpm --filter @aiostreams/core --filter @aiostreams/server run build` passes.
- Booted the built server and exercised the parser directly:
  - 150KB JSON `PUT /api/v1/user` → passes the body parser (proceeds to auth validation) — previously 413.
  - 2MB JSON → still rejected by the parser, confirming the limit remains enforced.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -57,6 +57,11 @@ DATABASE_URI=sqlite://./data/db.sqlite
 # The port the addon listens on.
 # PORT=3000
 
+# Maximum request body size accepted by the server. Large configurations can
+# exceed Express' 100kb default, causing config saves to fail with
+# "request entity too large".
+# MAX_REQUEST_BODY_SIZE=1mb
+
 # Optionally override the internal URL used when communicating with built-in
 # addons. The default is fine for almost all setups.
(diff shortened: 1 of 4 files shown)


dreulavelle:
LGTM

PR #1199, opened 2026-08-13
feat(presets/torrentclaw): add TorrentClaw preset

lucaiarr:
Adds a preset for [TorrentClaw](https://torrentclaw.com), a torrent/debrid aggregator that merges several indexers, deduplicates by info hash and scores each release by quality.

It can already be added as a custom addon today — but its stream metadata does not survive the trip, and that is what this preset fixes.

## The problem

The base parser derives the service — and with it `cached` — from `parseServiceData(stream.name)`, which looks for a provider name ("RD", "TorBox", …) in the stream name.

TorrentClaw never puts one there. The provider comes from the user's account, so the name only carries the playback state and the quality tier: `⚡️ INSTANT 1080p`, `⬇️ Download 1080p`. The provider regex never matches, `getService` returns `undefined`, and three things follow:

| | today | why it matters |
|---|---|---|
| `sorter.ts:81` | a service-less stream counts as cached | cache-on-demand entries sort in among the instant ones |
| `streamExpression.ts:1011` `cached()` | tests `=== true` → excluded | the genuinely cached entries are dropped |
| `getStreamType:534` | falls through to `http` | never classified as `debrid` |

Measured over **244 real streams** from the addon: **0 get a service today**, so 0 get a correct cache state. With the preset, every debrid entry is attributed and its cache state is correct, and the `⚡ INSTANT` entries are classified `debrid` instead of `http`.

## How it works

The parser reads flags the addon publishes instead of scraping the title text:

- `behaviorHints.cached` — the cache state, set from a verified provider check
- `behaviorHints.tcSource` — `p2p` / `debrid` / `usenet` / `library`, the delivery mechanism. Needed because everything served over HTTPS looks alike to a parser: a `url` with no `infoHash`
- `behaviorHints.tcAction` — present only on the non-playable cards

**Sizes need no code at all**: the addon sends `behaviorHints.videoSize` / `folderSize`, which the base parser already prefers over parsing text. That part works today for anyone using TorrentClaw as a custom addon.

## Details that are deliberate

Each of these was a bug in an earlier revision, caught by testing against real responses:

- **The service fallback applies only to `tcSource: "debrid"`.** A `service` means "this provider served the stream", so attaching one to a swarm entry is false — and `filterer.ts:2046` acts on it: `excludeUncached` drops anything with `service.cached === false`. Attributing the P2P entries deleted 50 of 80 streams that survive with the base parser.
- **Non-playable cards get no service at all.** The base heuristic scans the name for provider names and hits ordinary prose: the Spanish *"Ver **EN** tu navegador"* matches Easynews' known name `EN`, which then made that card vanish under `excludeUncached`. The preset is better than the base here.
- **`usenet` is only claimed when a service is attributed.** `DeduplicatorOptions` has no `usenet` key, so a usenet stream that is not re-labelled to cached/uncached resolves to an undefined mode, matches no branch of the switch, and is dropped silently.
- **`p2p` is only claimed when there is an `infoHash`.** The Stremio transformer passes `infoHash` and drops `url` for that type, so a p2p stream without one would reach the client with neither and play nothing.
- **`tcSource: "library"` maps to `getInLibrary`** — the file is already on the user's machine, which is what the sorter and the deduplicator's library tiebreaker are for.

## Why `generateAddons` returns a single addon

Unlike most debrid presets. Comet and friends put each provider's credentials in the manifest URL, so one instance per service is one distinct endpoint per service. TorrentClaw resolves the provider from the account behind the API key and exposes no provider field a third party can set, so **every instance built from the same options yields a byte-identical URL** — N instances would be N copies of one request, duplicating every result, not just the P2P ones.

For the same reason there is no `includeP2P`: P2P results always come back alongside the debrid ones in the same response, so the option is `excludeP2P` and it filters in the parser rather than issuing another request.

## Testing

Verified against the live addon, not fixtures alone:

- Generated manifest URLs resolve and return a valid manifest. An empty config is **not** encoded — the 3-character `e30` segment is not recognised as a config blob and answers `{"error":"Unknown endpoint"}`.
- The compiled parser was run over real responses: anonymous, debrid-enabled, a season pack, and Spanish copy. **No exceptions raised, and no stream lost relative to the base parser** (`excludeUncached` removes the same count or fewer).
- Degenerate inputs (missing `behaviorHints`, non-boolean `cached`, unknown `tcSource`, absent `url` and `infoHash`) degrade to the base behaviour rather than throwing.
- Addons that do not send the new fields — including older TorrentClaw versions — parse exactly as they do today.

`pnpm -F core build` is clean and the new files are Prettier-formatted. `pnpm gen:env-docs` was run for the three new environment variables.

Happy to adjust anything — naming, option defaults, or the single-instance decision if you'd rather it followed the usual per-service shape.

## Summary by CodeRabbit

* **New Features**
  * Added TorrentClaw as a supported preset.
  * Added configuration for the service URL, request timeout, and User-Agent.
  * Added stream handling for cache status, service attribution, delivery type, library status, action cards, and peer-to-peer results.
  * Added support for API key and language settings when generating addon links.

* **Documentation**
  * Documented TorrentClaw runtime configuration options.

### DIFF
--- a/packages/core/src/config/schema/presets.ts
+++ b/packages/core/src/config/schema/presets.ts
@@ -155,6 +155,13 @@ export const presetsSchema = {
       },
     },
   }),
+  torrentclaw: basicPreset({
+    label: 'TorrentClaw',
+    default: ['https://addon.torrentclaw.com/api/stremio'],
+    envBase: 'TORRENTCLAW_URL',
+    timeoutEnv: 'DEFAULT_TORRENTCLAW_TIMEOUT',
+    userAgentEnv: 'DEFAULT_TORRENTCLAW_USER_AGENT',
+  }),
   meteor: basicPreset({
     label: 'Meteor',
     default: ['https://meteorfortheweebs.midnightignite.me'],
--- a/packages/core/src/presets/presetManager.ts
+++ b/packages/core/src/presets/presetManager.ts
@@ -79,6 +79,7 @@ import { StreamNZBPreset } from './streamnzb.js';
 import { DavexPreset } from './davex.js';
 import { HdHubPreset } from './hdhub.js';
 import { BaguettioPreset } from './baguettio.js';
+import { TorrentClawPreset } from './torrentclaw.js';
 import { Preset } from './index.js';
 
 let PRESET_LIST: string[] = [
@@ -90,6 +91,7 @@ let PRESET_LIST: string[] = [
   'comet',
   'meteor',
   'mediafusion',
+  'torrentclaw',
   'stremthruTorz',
   'stremthruStore',
   'sootio',
@@ -195,6 +197,8 @@ export class PresetManager {
         return MeteorPreset;
       case 'mediafusion':
         return MediaFusionPreset;
+      case 'torrentclaw':
+        return TorrentClawPreset;
       case 'custom':
         return CustomPreset;
       case 'torbox':
--- a/packages/core/src/presets/torrentclaw.ts
+++ b/packages/core/src/presets/torrentclaw.ts
@@ -1,0 +1,421 @@
+import { Addon, Option, ParsedStream, Stream, UserData } from '../db/index.js';
+import { baseOptions, Preset } from './preset.js';
+import { constants, ServiceId } from '../utils/index.js';
+import { config as appConfig } from '../config/index.js';
+import { StreamParser } from '../parser/index.js';
+
+/**
+ * TorrentClaw publishes machine-readable stream metadata in `behaviorHints`, so
+ * this parser reads flags instead of scraping the human-facing title text.
+ *
+ * The base parser derives the service (and with it `cached`) from
+ * `parseServiceData(stream.name)`, which looks for a provider name — "RD",
+ * "TorBox", … — in the stream name. TorrentClaw never names the provider there:
+ * the user picks the debrid provider in their own addon config, so the name only
+ * carries the playback state and the quality tier ("⚡️ INSTANT 1080p",
+ * "⬇️ Download 1080p"). The provider regex therefore never matches, the service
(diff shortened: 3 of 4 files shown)


Cxsmo-ai:
My fork of aiostreams has torrentclaw integrated as best as I could fixed formating even usenet

lucaiarr:
Rebased onto current `main` — the branch was 231 commits behind. New head: `37d9a6de`. No conflicts; `pnpm install --frozen-lockfile` and `pnpm run build` both pass locally on Node 24.

@Viren070 a heads-up on the red checks, in case they've been putting you off looking at this: **CI has never actually run on this PR.** The four runs from 13 Aug report `conclusion: failure`, but they have `total_count: 0` jobs and their logs 404. They were queued as `action_required` (fork workflow approval), sat there untouched, and GitHub expired them exactly 30 days later on 12 Sep — which is when the API shows them flipping to failure. The "This run likely failed because of a workflow file issue" banner is generic and misleading here; `build.yml` and `release.yml` at that SHA are unmodified and valid.

The runs on the new head are in the same `action_required` state. Whenever you have a moment, approving the workflow runs would let the build actually report.

For reference, the change is self-contained — 3 source files, no existing behaviour touched:

- `packages/core/src/presets/torrentclaw.ts` — new preset + a `StreamParser` subclass that reads `behaviorHints` (`cached`, `tcSource`, `tcAction`) instead of scraping the title, since TorrentClaw never names the debrid provider in the stream name.
- `packages/core/src/presets/presetManager.ts` — registration.
- `packages/core/src/config/schema/presets.ts` — `basicPreset` entry.
- Plus the env-var docs table.

Happy to rebase again or split anything out if you'd prefer a smaller first cut.

PR #1200, opened 2026-08-13
feat(templates): ask before a template update overwrites your changes

IbbyLabs:
## The problem

Apply a template, then change your sort order to suit how you watch. When the
template publishes an update and you apply it, your sort order is silently
replaced by the template's, with no prompt and no sign that anything of yours was
overwritten.

This isn't a bug in how templates apply — the template legitimately sets
`sortCriteria`, and applying it does what it says. The problem is that applying an
update can't tell a setting the user deliberately changed from one they never
touched, so it replaces both. `debrid-starter.json` sets 51 top-level keys, so an
update potentially replaces 51 settings when the user had an opinion about one.

## What this does

Where an update and the user have both moved the same setting, the wizard asks
which to keep. Where only one of them moved it, nothing is asked and the update
applies as before.

```
user changed sorting, update changes sorting and resolutions
  -> asks about sortCriteria only

user changed nothing                -> asks nothing
update changes nothing they touched -> asks nothing
no stored base                      -> asks about every changed field, once
```

The middle two rows are what make it usable. An update touching forty settings the
user has no opinion about still goes through silently.

## How it works

`appliedTemplates[]` gains the template's config as it was applied. That is the
missing piece — with `id` and `version` alone there is nothing to compare against,
so a change of the user's and a change of the template's are indistinguishable.

With a base stored there are three things to compare, and no setting has to record
where it came from:

| | |
|---|---|
| base | the template's config as applied |
| mine | the user's config now |
| theirs | the template at the new version |

`base vs mine` is what the user changed. `base vs theirs` is what the update
changes. The overlap is the only thing worth asking about.

With no base — anyone who applied a template before this lands — every field the
update would change is treated as a conflict. Noisy once, and correct: assuming
the user changed nothing is how settings get overwritten.

The snapshot is taken before the wizard writes any service credentials into the
config, and `templateSnapshot` empties `services[].credentials` regardless, so
nothing of the user's ends up stored in `appliedTemplates[]`.

**Try it** — multi-arch image built from this branch:
```
ghcr.io/ibbylabs/aiostreams:pr1200-e615fc1d
```
Pinned to commit `e615fc1d`, so this tag keeps pointing at what was reviewed even if the branch moves. Unofficial build from an unmerged branch, published under `ibbylabs` rather than by the maintainer.

This branch adds a field to `UserDataSchema` but no database migration, so switching back to your usual image is clean.

## Summary by CodeRabbit

- **New Features**
  - Added conflict detection when updating configuration templates.
  - Users can review differences between their current settings and incoming changes.
  - Added options to keep individual settings, apply updates, or cancel.
  - Template updates preserve user-modified values and existing credentials where appropriate.
  - Sensitive service credentials are excluded from saved template snapshots.
- **Bug Fixes**
  - Improved handling of template settings and service metadata during updates.
  - Preserved service metadata while protecting credentials in configuration snapshots.

### DIFF
--- a/packages/core/src/db/schemas.ts
+++ b/packages/core/src/db/schemas.ts
@@ -537,6 +537,11 @@ export const UserDataSchema = z.object({
         id: z.string(),
         version: z.string(),
         url: z.string().optional(),
+        // The template's config as it was applied, so an update can be merged
+        // against what the user started from rather than overwriting it. Never
+        // contains credentials: it is captured before the user's inputs are
+        // stamped in.
+        config: z.record(z.string(), z.unknown()).optional(),
         dismissedVersion: z.string().optional(), // dismissed update notification up to this version
         ignored: z.boolean().optional(), // permanently ignore all future update notifications
       })
--- a/packages/frontend/src/components/shared/templates/conflict-step.tsx
+++ b/packages/frontend/src/components/shared/templates/conflict-step.tsx
@@ -1,0 +1,86 @@
+'use client';
+
+import { useState } from 'react';
+import { Button } from '@/components/ui/button';
+import { Checkbox } from '@/components/ui/checkbox';
+import type { TemplateUpdateConflict } from '@/lib/templates/merge';
+
+interface Props {
+  conflicts: TemplateUpdateConflict[];
+  isLoading?: boolean;
+  onCancel: () => void;
+  onConfirm: (keepMine: string[]) => void;
+}
+
+/**
+ * Shown when a template update would change a setting the user has changed
+ * themselves. Only those settings appear: anything the user never touched is
+ * updated without asking.
+ */
+export function TemplateConflictStep({
+  conflicts,
+  isLoading,
+  onCancel,
+  onConfirm,
+}: Props) {
+  // Keeping the user's own values is the default. The update is what they asked
+  // for, but their changes are the thing they would notice losing.
+  const [keepMine, setKeepMine] = useState<Set<string>>(
+    () => new Set(conflicts.map((c) => String(c.field)))
+  );
+
+  const toggle = (field: string, keep: boolean) => {
+    setKeepMine((prev) => {
+      const next = new Set(prev);
+      if (keep) next.add(field);
+      else next.delete(field);
+      return next;
+    });
+  };
+
+  return (
+    <div className="flex flex-col gap-4 overflow-y-auto">
+      <p className="text-sm text-[--muted]">
+        These settings differ from the template because you changed them. Tick
+        the ones to keep as they are; the rest will take the update. Everything
+        else in this template updates either way.
+      </p>
+
+      <div className="flex flex-col gap-2">
+        {conflicts.map((conflict) => {
+          const field = String(conflict.field);
+          return (
+            <label
+              key={field}
+              className="flex items-center gap-3 rounded-[--radius] border border-[--border] p-3"
+            >
+              <Checkbox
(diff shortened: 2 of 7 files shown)


IbbyLabs:
Two of the three are addressed in `f854b6a2`.

The completion actions now live in one `finish()` called by both the immediate apply and the deferred one, so a template that sets `setToSaveInstallMenu` still switches menus when the user had a conflict to resolve — it previously did not. `cancelConflicts` now clears the same state `handleCancel` does.

On recursive path-level diffing: not doing it, and it's a deliberate boundary rather than an oversight.

Of the 51 top-level keys the shipped `debrid-starter.json` sets, 38 are ordered arrays where the order *is* the setting — `preferredResolutions` is a preference ranking, and `sortCriteria.global` is an array of `{key, direction}` whose position is the sort precedence. Diffing those by path produces "you changed element 3", which isn't a question a user can answer. It would add precision only on the 11 object-valued keys, and `sortCriteria` is an object whose one child is an array, so descending a level lands back on an array there too.

The cost is real and worth stating: if you keep your `sortCriteria`, you also discard any unrelated change the update made inside it. That's the trade, and it's the same granularity the config already uses everywhere else.

IbbyLabs:
@coderabbitai review

IbbyLabs:
@coderabbitai full review

IbbyLabs:
Both addressed.

`services` are now merged into a new array inside the updater rather than assigned back into a variable from the enclosing scope — an updater has to be safe to call twice, and StrictMode does exactly that. Behaviour is the same; the side effect is gone.

`pendingCommit` now takes `string[]`, which is what every caller passes.

IbbyLabs:
@coderabbitai full review

IbbyLabs:
Good catch, and the second half of it was the important part.

A field the template still sets at the value it always set was being reported as a conflict whenever the user had changed it — the update hadn't touched it, so it shouldn't have prompted. And as you say, suppressing the prompt alone would have been worse: `commit` applied the whole config, so that field would have been reset silently instead.

The planner now compares the incoming config against the stored snapshot first. A field the update leaves alone isn't part of the update, so it neither prompts nor overwrites, and `commit` applies only the changed set minus anything kept. With no snapshot everything still applies, which is the existing behaviour for a first apply.

```
update leaves the field alone, user changed it   not prompted, not overwritten
update changes it, user changed it too           prompted
update changes it, user did not                  applied silently
no snapshot                                      everything applies, as before
```

IbbyLabs:
One thing in that last change worth calling out explicitly, because it is a semantic change rather than a bug fix, and it is yours to accept or reject.

Applying a template update now means "apply what the template changed" rather than "reapply the template". A field the template still sets at the value it always set is left alone entirely — not prompted, not written.

The alternative is to keep reapplying the whole config and prompt whenever the user has customised any field the template sets, changed or not. That is safe, but it asks about settings the update is not changing, and a prompt that fires on things that are not changing is one people learn to click through.

Only the delta path does this. With no stored snapshot — a first apply, or a template applied before this existed — everything still applies exactly as it does today.

Happy to drop it back to the whole-config apply if you would rather the update keep its current meaning; it is contained to the planner and the commit payload.

IbbyLabs:
Opened #1209 to track the underlying problem separately, so it's visible without reading the diff. Happy to close either if you'd rather it lived in one place.

IbbyLabs:
Superseded by 80dd827c7f8e35b44ea6bb8c6564ea040b29def7

PR #1210, opened 2026-08-16
feat(failover): detect debrid error videos and fail over to the next stream

javi11:
## Problem

When a debrid link is dead or unavailable, providers don't return an HTTP error — they return a short "content unavailable" video. `service.resolve()` hands that link back without complaint, so `resolvePlaybackTarget` treats it as a success, it wins the play chain, and the user gets a few seconds of error video with no failover.

The heuristics for spotting one already exist in this repo, but only for *external* addon URLs, inline in `resolveExternalTarget` (16 MiB floor, content-type gate, same-host-redirect rule). The owned debrid path — the main one — never probed anything.

## Change

Extract those heuristics into `classifyProbedBody` (pure, unit-tested) in a new `packages/core/src/main/playback-probe.ts`, and reuse them on the owned path behind a new opt-in `failover.verifyPlayback` option.

When enabled, the resolved URL is probed with a `Range: bytes=0-0` GET before being served. If it looks like an error video, the attempt throws a plain `Error` — no debrid `code`, so `isFailoverRetryableError` treats it as retryable and the chain advances to the next stream.

`resolveExternalTarget` is refactored onto the same classifier. Its redirect handling is untouched and it keeps its existing semantics.

### On sizing the check

I initially compared the served size against the size the source advertised, which would catch error clips above the 16 MiB floor. I dropped it: `ParsedStream.size` is only reliably per-file for the builtins — `getSize()` otherwise falls back to whatever an external addon put in `stream.size` or parsed out of its description, which is frequently a season-pack total. A 2 GiB episode from a release advertised at 20 GiB would have been rejected as an error video. The blanket floor is enough in practice, since provider error clips are ~1–10 MB.

### Drive-by fix

The probe timeout never applied. `makeRequest` does `options.signal ?? AbortSignal.timeout(options.timeout)`, and failover always passes a per-attempt signal, so both probes were bounded only by the whole chain deadline (60s default) rather than 10s. They now race the caller's signal against their own timeout via `AbortSignal.any`, so a hanging CDN fails fast instead of stalling an attempt for a minute.

## Behaviour

- **Off by default** (`DEFAULT_FAILOVER_VERIFY_PLAYBACK = false`). No change for anyone who doesn't enable it.
- Costs one extra request per attempt, so playback starts slightly slower. Called out in the UI help text and the docs.
- The clicked item is only verified when fallbacks exist — with nothing to advance to, rejecting it just trades an error video for an error page, so there's no upside to risking a false positive.
- Play chains cached before this change default `verifyPlayback` to `false`, so nothing breaks across a deploy.
- The schema addition is a plain optional boolean; no migration needed on either driver.

## Testing

15 unit tests in `packages/core/src/main/playback-probe.test.ts` covering the classifier (206 vs 200, missing `content-range`, content-type gate, floor boundary, non-success status) and `verifyPlaybackUrl` (body cancellation, retryable error shape, request shape, signal/timeout composition).

`pnpm -F core test` 15/15, `core` + `server` build, `frontend` typecheck and lint clean.

## Files

- **new** `packages/core/src/main/playback-probe.ts`, `packages/core/src/main/playback-probe.test.ts`
- `packages/core/src/main/failover.ts` — shared classifier, `ctx.verify` on `resolvePlaybackTarget`, timeout fix
- `packages/core/src/main/play-chain.ts`, `resources.ts`, `db/schemas.ts`, `utils/constants.ts`, `index.ts` — flag plumbing
- `packages/server/src/routes/api/debrid.ts` — wiring
- `packages/frontend/src/components/menu/services/_components/builtin-settings.tsx` — toggle
- `packages/docs/content/docs/guides/usenet.mdx` — documented under Failover → Key Settings

## Summary by CodeRabbit

* **New Features**
  * Added an optional **Verify Playback Before Serving** failover setting.
  * Resolved playback links can be checked for unavailable-content responses before redirection.
  * Invalid links are rejected so failover can continue to the next available result.
  * Added playback verification across playback resolution and failover processing, disabled by default.
* **Documentation**
  * Documented the setting, including its additional verification request per attempt.
* **Tests**
  * Added coverage for response validation, timeouts, cancellation, retries and range requests.

### DIFF
--- a/packages/core/src/db/schemas.ts
+++ b/packages/core/src/db/schemas.ts
@@ -901,6 +901,8 @@ export const UserDataSchema = z.object({
       precacheFailover: z.boolean().optional(),
       /** Include non-owned addon debrid URLs (resolved by probing) as failover targets. Default false. */
       includeExternalFailover: z.boolean().optional(),
+      /** Probe each resolved playback URL and fail over when the debrid service serves a short error video instead of the release. Costs one extra request per attempt. Default false. */
+      verifyPlayback: z.boolean().optional(),
       /** Max same-release variant attempts tried per release before moving on (0 disables). */
       sameReleaseLimit: z.number().min(0).optional(),
       /** Delay between launching same-release variant attempts (ms). Default 0. */
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -69,3 +69,15 @@ export type {
   RunPlayChainConfig,
   RunPlayChainResult,
 } from './main/failover.js';
+export {
+  classifyProbedBody,
+  createProbeSignal,
+  verifyPlaybackUrl,
+  MIN_PLAUSIBLE_FILE_SIZE,
+  DEFAULT_PROBE_TIMEOUT_MS,
+} from './main/playback-probe.js';
+export type {
+  ProbeVerdict,
+  ProbedBody,
+  VerifyPlaybackOptions,
+} from './main/playback-probe.js';
--- a/packages/core/src/main/failover.ts
+++ b/packages/core/src/main/failover.ts
@@ -17,6 +17,12 @@ import {
   PLAYBACK_PATH_PREFIX,
 } from '../debrid/utils.js';
 import { isFailoverRetryableError } from './play-chain.js';
+import {
+  classifyProbedBody,
+  createProbeSignal,
+  verifyPlaybackUrl,
+  DEFAULT_PROBE_TIMEOUT_MS,
+} from './playback-probe.js';
 
 const logger = createLogger('failover');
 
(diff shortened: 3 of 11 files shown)


Viren070:
debrid errors from services in AIOStreams return a DebridError though, which service returns an error video?

javi11:
This kind of video image

Viren070:
This looks like an error video served by Torrentio. When you open a Torrentio stream, you go to a Torrentio URL and it handles resolving and then redirecting to their own error video if it fails. The service resolving code in AIOStreams can't affect this case since you would never even reach the AIOStreams `/api/v1/debrid` endpoint when your client is playing a stream from an external addon. 

The external addon option in failover settings is so that you can failover from a built-in addon/service wrapped addon to an external addon but if you click on an external addon stream first, AIOStreams can't control what happens there. 

You'd need to service wrap Torrentio so that AIOStreams becomes the one to resolve the torrentio streams and then aiostreams failover can work for it.

javi11:
Oh make sense didn't knew that. Thanks, closing the pr

PR #1235, opened 2026-08-24
feat(core): parse StremThru's edition data, fix emoji-boundary regex

webstreamr:
getRegexForTextAfterEmojis missed emoji that only render via a variation selector (e.g. ⚙️), silently swallowing whatever followed them into the match. Extracted it into parser/utils.ts, fixed the boundary check, and updated every preset call site; added the first core test suite to cover it.

---

stremthru got the edition field with https://github.com/MunifTanjim/stremthru/pull/582

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved parsing of text following emoji markers, including variation-selector emojis, line endings and multiple supported emojis.
  - Improved extraction of audio, subtitle, release group, folder, message, proxy and error details from stream information.
  - Edition information following the 🏷️ marker is now correctly captured for file merging.
- **Improvements**
  - More consistent emoji-based metadata detection across supported stream sources.
  - Enhanced handling of end-of-line and end-of-text boundaries during parsing.

### DIFF
--- a/packages/core/src/parser/index.ts
+++ b/packages/core/src/parser/index.ts
@@ -1,2 +1,3 @@
 export { default as FileParser } from './file.js';
 export { default as StreamParser } from './streams.js';
+export { getRegexForTextAfterEmojis } from './utils.js';
--- a/packages/core/src/parser/streams.test.ts
+++ b/packages/core/src/parser/streams.test.ts
@@ -1,0 +1,37 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { getRegexForTextAfterEmojis } from './utils.js';
+
+describe('getRegexForTextAfterEmojis', () => {
+  it('stops at a default-presentation emoji', () => {
+    const re = getRegexForTextAfterEmojis(['🎞️']);
+    assert.equal('🎞️ HEVC\n🎧 DD'.match(re)?.[1], 'HEVC');
+  });
+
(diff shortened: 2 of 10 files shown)


Viren070:
Isn't the edition field just parsed from the filename?

webstreamr:
> Isn't the edition field just parsed from the filename?

StremThru sets it here https://github.com/MunifTanjim/stremthru/blob/e97392a78210d4b51cc556659e1c90a83d1f4a55/internal/torrent_info/db.go#L295 from the title using https://github.com/MunifTanjim/go-ptt it seems

AIOS gets it from the file name, right?

But since this is an override it can cause more issues than it solves, right? I guess it's better to close this then, not worth it. But maybe you can confirm please.

webstreamr:
or just keep the emoji stuff + tests :D

Viren070:
Yeah we both get it from the filename and anything StremThru is able to parse we should be able to as well since our parser is based off go-ptt (https://github.com/MunifTanjim/go-ptt/blob/d68f7eb2d535de515f3275885010d837d30a12c7/handlers.go#L701-L782 and https://github.com/Viren070/parse-torrent-title/blob/c83bdb8ecaff4d9e3fc942598cfc8ce46c857035/src/handlers.ts#L361-L477) so it should be unnecessary,

yeah we can keep emoji fix

webstreamr:
--> https://github.com/Viren070/AIOStreams/pull/1238

PR #1264, opened 2026-09-02
feat(core,frontend): expose stream url/urlHost to custom formatter + allow HTTP streams as probed failover targets

Qiiks:
# feat(core,frontend): expose stream url/urlHost to custom formatter + allow HTTP streams as probed failover targets

> **Scope note:** this PR was originally mixed with filename-residue cleanup, an adaptive resolution floor, and Docker cache mounts. Those were split into their own PRs: #1265 (residue), #1266 (adaptive floor), #1267 (docker caches). This PR now contains only the HTTP-failover + formatter-URL-fields feature, with all review findings addressed.

## What

Two related improvements for HTTP (direct-URL) streams:

### 1. Custom formatter: `{stream.url}` and `{stream.urlHost}`

Adds `url` and `urlHost` to the formatter's `stream` field registry. HTTP addons often serve the same release from several mirrors; users can now label or filter on the serving host, e.g.:

    {stream.resolution}{stream.urlHost::exists[" • {stream.urlHost}"||""]}

### 2. HTTP mirrors as probed failover targets

Same-file HTTP mirrors (identical `behaviorHints.filename`) previously could not participate in failover at all:

- the deduplicator's external-failover harvest gate (`isExternalFailoverTarget`, formerly `isExternalDebridFailover`) only accepted `type === 'debrid'`
- `FailoverContentType` had no `http` variant, so HTTP streams could not appear in a play chain

This PR:

- widens `FailoverContentType` with `'http'` (type-faithful end-to-end - labels show `http variant`, content-type filters behave)
- admits `http` streams into `isExternalFailoverTarget` (still restricted to the addon's own manifest host; owned playback URLs still excluded)
- keeps the true stream type through the harvest path instead of laundering mirrors as `debrid`
- accepts `'http'` in the `failover.contentTypes` and `failoverVariants.type` zod enums and adds an HTTP option to the failover content-types UI
- updates the Failover card description and the "Include External Addon Targets" help text to describe HTTP targets

The resolve path needs no changes: `resolveExternalTarget` already probes any URL with `Range: bytes=0-0`, redirect-host heuristics, and the 16 MiB plausibility floor, and the debrid route branches on `item.kind` (external -> probe), not on the content type.

Behavior parity with existing external debrid failover: mirrors become **failover targets** (tried when a clicked debrid/usenet item fails, when included in content types / cross-type), but a *direct click* on an HTTP stream still plays it directly without failover - same as external debrid links today.

## Review fixes in this update

- renamed `isExternalDebridFailover` -> `isExternalFailoverTarget` (predicate now admits both kinds)
- the contentTypes test now parses `UserDataSchema.pick({ failover: true })` instead of asserting on a cast object
- the chain test is a full round trip: 2+ eligible streams -> build -> decode stamped fallback key -> `getPlayChain` -> assert the external HTTP target resolves with `type: 'http'` (the old version passed one stream and hit buildPlayChain's <2-eligible early return)
- Failover UI help text updated for HTTP targets

## Why

HTTP addons (PenguPlay, ShowBox, Yukistreams, ...) frequently return 2-5 byte-identical mirrors of one release. With `deduplicator.http = 'single_result'` the duplicates collapse, but the dropped mirrors were pure waste before this change; now they serve as same-release fallbacks, probed for liveness at resolve time.

## Testing

- `pnpm build` clean in core, server, frontend
- 15 tests in `play-chain.test.ts` + `url-fields.test.ts`: widened guards, real zod enum parsing, full chain round trip with `http` items and variants, formatter field rendering incl. conditional `::exists` branches
- `pnpm test` in core: 40/40 passing
- deployed on a real instance: same-file mirror duplicates collapse under `http: single_result`, and HTTP mirrors appear as `http` variants in the stored play chain

### DIFF
--- a/packages/core/src/db/schemas.ts
+++ b/packages/core/src/db/schemas.ts
@@ -895,7 +895,7 @@ export const UserDataSchema = z.object({
     .object({
       enabled: z.boolean().optional(),
       /** Which result kinds may appear as failover targets. Default ['usenet']. */
-      contentTypes: z.array(z.enum(['usenet', 'debrid'])).optional(),
+      contentTypes: z.array(z.enum(['usenet', 'debrid', 'http'])).optional(),
       /** Allow a click on one kind to fail over into a different kind. Default false. */
       allowCrossType: z.boolean().optional(),
       /** Max total failover attempts (after de-duplication) tried after the clicked item. */
@@ -1211,7 +1211,7 @@ export const ParsedStreamSchema = z.object({
     .array(
       z.object({
         url: z.string(),
-        type: z.enum(['usenet', 'debrid']),
+        type: z.enum(['usenet', 'debrid', 'http']),
         serviceId: z.string().optional(),
         filename: z.string().optional(),
         identity: z.string().optional(), // nzbUrl | infoHash | external host+path
--- a/packages/core/src/formatters/base.ts
+++ b/packages/core/src/formatters/base.ts
@@ -6,6 +6,15 @@ import { compileTemplate as engineCompileTemplate } from './engine/compile.js';
 import { NEW_LINE_SENTINEL, REMOVE_LINE_SENTINEL } from './engine/sentinels.js';
 import { comparatorFunctions } from './engine/comparators.js';
 
+/** Host of a parseable URL, or null — used for the formatter's urlHost field. */
+function extractUrlHost(url: string): string | null {
+  try {
+    return new URL(url).host;
+  } catch {
+    return null;
+  }
+}
+
 /**
  *
  * The custom formatter code in this file was adapted from https://github.com/diced/zipline/blob/trunk/src/lib/parser/index.ts
@@ -47,6 +56,8 @@ export interface ParseValue {
   stream?: {
     filename: string | null;
     folderName: string | null;
+    url: string | null;
+    urlHost: string | null;
     size: number | null;
     bitrate: number | null;
     folderSize: number | null;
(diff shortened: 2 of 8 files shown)


Qiiks:
Thanks @Copilot @coderabbitai — all actionable findings are addressed in da8cb2c:

**Fixed here:**
- `play-chain.test.ts` (schema): now parses `UserDataSchema.pick({ failover: true })` — enum regressions fail the suite instead of a cast-object no-op. (Copilot + CodeRabbit)
- `play-chain.test.ts` (chain): rebuilt as a full round trip — 3 eligible streams, build, decode the stamped fallback key from the owned URL, resolve via `getPlayChain` with cross-type on, assert the external HTTP mirror resolves with `type: 'http'`. The previous version passed one stream and hit the <2-eligible early return, exercising nothing. (Copilot + CodeRabbit)
- `play-chain.ts` + call sites: renamed `isExternalDebridFailover` → `isExternalFailoverTarget` (predicate admits debrid + http). (Copilot)
- `builtin-settings.tsx`: Failover card description and "Include External Addon Targets" help now describe HTTP targets and their external-target requirement. (CodeRabbit)
- Dockerfile: production `pnpm install --prod` now mounts the same `/pnpm/store` cache. (CodeRabbit)

**Split out (scope):** filename-residue cleanup → #1265, adaptive resolution floor → #1266, Docker cache mounts → #1267. The residue regex tightening and the settings-card id landed in those PRs.

Copilot's suppressed note about the type-guard predicate narrowing: taken further — the rename plus the existing `s is ParsedStream & { url: string }` guard keeps narrowing faithful; callers no longer need re-mapping.

PR #1265, opened 2026-09-02
fix(parser): strip provider-appended media-info residue from HTTP stream filenames

Qiiks:
# fix(parser): strip provider-appended media-info residue from HTTP stream filenames

## What

Some HTTP providers (4KHDHub, VAPlayer, 2Peckle, CineFreak) glue a media-info echo onto the real release filename after the extension, e.g.:

```
Reacher.S01E01.1080p.WEB-DL.DDP5.1.H.264-NTb.mkv1080pMP4WEB-DLx264~9.5Mbps1080pmp4web-dlx264~9.5mbps.pad-2Peckle
```

That polluted `behaviorHints.filename` leaks into custom formatter output and into dedup keys, producing garbled stream descriptions like `26.2 Mbps4kmkvblurayhdr10+10`.

`stripProviderResidue` (new module `packages/core/src/parser/providerResidue.ts`) removes the residue:

- **Extension present**: takes the LAST media-extension candidate and cuts there when the trailing residue carries a provider marker (a `~N.Mbps` bitrate echo or a `.pad-<tag>` suffix). The residue is glued directly to the extension (`.mkv4KMKV…`), so a negative-lookahead match cannot work.
- **Extension dropped**: some providers glue a hex id plus the residue straight onto the title (`Show.2022.<32-hex>1080pMKVWEB-DL…~5.8Mbps…pad-X`). When no extension candidate exists, the cut happens at the hex token — and only when a container/bitrate marker follows (a bare `~Director.Cut` suffix is NOT treated as residue).
- **Legit names are returned byte-identical.**

`StreamParser.getFilename` applies it, so formatters, dedup keys and parsed metadata all see the cleaned name.

## Testing

- 8 new tests (`filename-clean.test.ts`): all four real-world polluted patterns from a live instance, pad-only residue, hex-glue without extension, and four negative cases (legit names untouched, junk without markers kept, non-bitrate tilde suffix kept)
- full core suite green; deployed on a live instance — polluted rows now render as `Reacher.S01E01.1080p.AMZN.WEB-DL.DDP5.1.H.264-NTb.mkv`

Residue cleanup for HTTP stream filenames, split out of #1264 per scope.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved filename handling for media streams by removing recognised provider-generated metadata.
  - Preserved legitimate filenames and unrelated trailing text when no recognised provider markers are present.
  - Improved handling of filenames with missing extensions when identifiable provider metadata follows a recognised identifier.
  - Reduced false positives when processing `.pad-` suffixes by requiring relevant media information before removing them.
  - Applied consistent cleanup to filenames provided through stream metadata hints.

### DIFF
--- a/packages/core/src/parser/filename-clean.test.ts
+++ b/packages/core/src/parser/filename-clean.test.ts
@@ -43,6 +43,11 @@ describe('stripProviderResidue', () => {
       'Reacher.2022'
     );
   });
+  it('keeps a hex-id name whose tilde suffix has no bitrate marker', () => {
+    const name = 'Show.abcdefabcdefabcdef~Director.Cut';
+    assert.equal(stripProviderResidue(name), name);
+  });
+
 
   it('keeps a filename whose post-extension junk has no provider marker', () => {
     const name = 'Reacher.S01E01.mkvReacher fake no marker';
--- a/packages/core/src/parser/filename-clean.test.ts
+++ b/packages/core/src/parser/filename-clean.test.ts
@@ -47,6 +47,11 @@ describe('stripProviderResidue', () => {
     const name = 'Show.abcdefabcdefabcdef~Director.Cut';
     assert.equal(stripProviderResidue(name), name);
   });
+  it('keeps a legit ".pad-" suffix that has no media-info echo before it', () => {
+    const name = 'Movie.mkv.pad-notes';
+    assert.equal(stripProviderResidue(name), name);
+  });
+
 
 
   it('keeps a filename whose post-extension junk has no provider marker', () => {
--- a/packages/core/src/parser/filename-clean.test.ts
+++ b/packages/core/src/parser/filename-clean.test.ts
@@ -32,7 +32,16 @@ describe('stripProviderResidue', () => {
 
   it('keeps a legit filename untouched', () => {
     const name = 'Reacher.S01E01.1080p.WEB.h264-KOGi.mkv';
+
     assert.equal(stripProviderResidue(name), name);
+  });
+  it('cuts hex-id glue when the provider dropped the extension entirely', () => {
+    assert.equal(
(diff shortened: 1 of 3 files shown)


Qiiks:
Scope: split out of #1264 (filename-residue cleanup). Includes the CodeRabbit fix from the #1264 review: the hex-id glue lookahead now requires the bitrate marker (`~\d+(?:\.\d+)?mbps`) instead of a bare `~`, with a regression test (`Show.<hex>~Director.Cut` is kept intact). @coderabbitai review

Qiiks:
Good catch — fixed in 69ea8c3. `RESIDUE_MARKER` now requires a media-info echo (resolution/container/codec token) before `.pad-` instead of accepting it bare: `/(?:~\d+(?:\.\d+)?mbps)|(?:(?:mkv|mp4|avi|webm|bluray|web-?dl|x264|x265|hevc|av1|\d{3,4}p)[a-z0-9.+-]*\.pad-)/i`. Added the negative test: `Movie.mkv.pad-notes` is returned unchanged, while `…mkv1080pMKVBluRayx264.pad-VAPlayer` (echo present) is still stripped. 9/9 tests green.

Qiiks:
Fixed in 69ea8c3→(extension list): `MEDIA_EXTENSIONS` now mirrors the full container set the parser already recognizes in `streams.ts` (3gp, 3g2, vob, ogv, ogm, divx, xvid, rm, rmvb, asf, mxf, mka, mks, mk3d, f4v/f4p/f4a/f4b, m2ts, ts) so residue on those containers is stripped and the two lists can't silently diverge. 9/9 tests green.

Qiiks:
@coderabbitai review

Viren070:
This was for pengu play specifically?

PR #1266, opened 2026-09-02
feat(core,frontend): adaptive resolution floor - show available streams when the floor is unmatched

Qiiks:
# feat(core,frontend): adaptive resolution floor

## What

`requiredResolutions` currently hard-removes every stream whose parsed resolution is not in the allowlist. On sources that don't declare a resolution (many Asian-drama/HTTP addons), that surfaces a **"Required Resolution (4) - 4x Unknown"** removal-reasons card and an empty player.

This adds an opt-in `userData.adaptiveResolutionFloor` flag (new toggle: **Filters → Resolution → Adaptive Resolution Floor**):

- Before the filter pass, the filterer checks whether **any** stream in the result set meets the floor.
- **At least one matches** → the floor is enforced as usual (only the allowed resolutions are shown).
- **Nothing matches** → the floor is dropped for that request and everything is shown, so the user sees what exists instead of nothing.

Per-request, set-aware, zero cost when the flag is off.

## Implementation

- `packages/core/src/db/schemas.ts`: `adaptiveResolutionFloor: z.boolean().optional()`
- `packages/core/src/streams/filterer.ts`: after `filterableStreams` is computed, one `streams.some(...)` decides a `bypassRequiredResolutions` flag that the existing `requiredResolutions` check consults; bypass logs an info line for traceability
- `packages/core/src/utils/fieldMeta.ts` + filters UI: field metadata and a labelled `SettingsCard` (id `adaptiveResolutionFloor` for command-palette scroll)

## Testing

- 4 new tests (`adaptive-floor.test.ts`): floor enforced when a match exists / floor dropped when nothing matches / hard-filter unchanged when flag off / mixed set keeps enforcement
- full core suite green (51/51); verified live: a 720p+-available title shows only 720p+ rows, a no-resolution drama shows all streams instead of a removal card

Split out of #1264 per scope.

## Summary by CodeRabbit

* **New Features**
  * Added an **Adaptive Resolution Floor** setting under Resolution filters.
  * When enabled, streams remain available if none meet the configured required resolution; the requirement is relaxed for that request.
  * Required resolution filtering continues to apply whenever at least one stream meets the configured requirement.
  * Resolution filtering is now evaluated across the complete set of streams for each request, ensuring consistent results across multiple sources.

### DIFF
--- a/packages/core/src/db/schemas.ts
+++ b/packages/core/src/db/schemas.ts
@@ -554,6 +554,8 @@ export const UserDataSchema = z.object({
   excludedResolutions: z.array(Resolutions).optional(),
   includedResolutions: z.array(Resolutions).optional(),
   requiredResolutions: z.array(Resolutions).optional(),
+  /** When no stream matches requiredResolutions, keep everything instead of showing nothing. */
+  adaptiveResolutionFloor: z.boolean().optional(),
   preferredResolutions: z.array(Resolutions).optional(),
   excludedQualities: z.array(Qualities).optional(),
   includedQualities: z.array(Qualities).optional(),
--- a/packages/core/src/streams/adaptive-floor.test.ts
+++ b/packages/core/src/streams/adaptive-floor.test.ts
@@ -55,9 +55,10 @@ describe('adaptiveResolutionFloor', () => {
   const floor = ['2160p', '1440p', '1080p', '720p'];
 
   // ':memory:' doesn't survive connect.ts's URL parsing and absolute
-  // Windows paths don't fit its URL scheme, so use a relative path in the
-  // package cwd; after() removes every artifact.
-  const dbPath = 'adaptive-floor-test.db';
+  // Windows paths don't fit its URL scheme, so use a relative path in
+  // the package cwd, unique per process so concurrent runs can't touch
+  // each other's files; after() removes every artifact.
+  const dbPath = `adaptive-floor-test-${process.pid}.db`;
   before(async () => {
     await initDb(`sqlite://./${dbPath}`);
     await initialiseConfig();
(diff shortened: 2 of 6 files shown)


Qiiks:
Scope: split out of #1264 (adaptive resolution floor). Includes the CodeRabbit fix from the #1264 review: the settings card now carries `id="adaptiveResolutionFloor"` so command-palette navigation can scroll to it. @coderabbitai review

Qiiks:
@coderabbitai full review

Qiiks:
All three findings fixed in d9fab0e:
- test DB: `after()` now calls `closeDb()` (releasing better-sqlite3's handle) then removes the `.db`/`-shm`/`-wal` artifacts — verified the workspace is clean after a run. Kept a relative path because `:memory:` doesn't survive `connect.ts`'s URL parsing and absolute Windows paths don't fit its URL scheme.
- the 'enforces the floor' assertion now states the expected result directly (`['720p']`).
- the conditional-type infer chain is replaced with `NonNullable<UserData['requiredResolutions']>[number]`.

Qiiks:
@coderabbitai review

Qiiks:
Both round-3 findings addressed in 9f9eded:

**Major (request-scope floor) — agreed and fixed.** `filter()` indeed runs once per fetch batch under dynamic addon fetching / addon groups, so the adaptive decision could not see later batches. The fetcher now runs one request-scope pass over the merged result of every batch via a new `filterer.applyResolutionFloor()`: if any stream anywhere in the request meets the floor, off-floor streams retained by earlier batches are retracted; if none does, everything stays. Regression test covers exactly the cross-batch case (batch 1 = 480p/Unknown, batch 2 = 1080p → only 1080p survives).

**Minor (fixed db name) — fixed.** `adaptive-floor-test-${process.pid}.db` keeps the relative path (required by connect.ts's URL scheme) but is unique per process, and `after()` still removes all artifacts.

Qiiks:
@coderabbitai review

PR #1267, opened 2026-09-02
build(docker): persist pnpm store and build cache across builds

Qiiks:
# build(docker): persist pnpm store and build cache across builds

## What

Every image build currently re-downloads all dependencies and rebuilds tooling artifacts from scratch. This mounts BuildKit cache volumes:

1. `pnpm store` at `/pnpm/store` for the dependency install (`pnpm config set store-dir /pnpm/store` inside the same `RUN`), so package tarballs download once and are reused across builds
2. `/build/node_modules/.cache` for the build step, so frontend bundler artifacts persist
3. the production `pnpm install --prod` also mounts the same store, so prod dependencies resolve from cache after `node_modules` is cleared (previously only the first install was cached)

Cache mounts never enter the image layers - they only speed up builds on the same builder, so the runtime image is unchanged.

## Testing

- `docker build` locally on linux/arm64: cold build unchanged, warm build skips the dependency download step
- deployed via CI-equivalent builds on a live instance

## Summary by CodeRabbit

* **Performance**
  * Improved container build efficiency by using stable, dedicated caches for package installation and build processes.
  * Reduced repeated downloads and processing during development and deployment builds.
* **Reliability**
  * Prevented concurrent builds from corrupting shared build caches.
  * Isolated development and pull request build caches from deployment workflow caches.

### DIFF
--- a/.github/workflows/dev-docker.yml
+++ b/.github/workflows/dev-docker.yml
@@ -96,8 +96,11 @@ jobs:
           context: .
           file: ./Dockerfile
           outputs: type=image,name=ghcr.io/viren070/aiostreams,push-by-digest=true,name-canonical=true
-          cache-from: type=gha,scope=${{ matrix.suffix }}
-          cache-to: type=gha,mode=max,scope=${{ matrix.suffix }}
+          # Trust separation: dev/PR builds must never share cache scopes with
+          # the deployment workflow, so poisoned cache data cannot flow into
+          # production images.
+          cache-from: type=gha,scope=dev-${{ matrix.suffix }}
+          cache-to: type=gha,mode=max,scope=dev-${{ matrix.suffix }}
 
       - name: Export digest
         run: echo "${{ steps.build.outputs.digest }}" > "/tmp/${{ matrix.suffix }}"
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,10 @@ RUN rm -rf packages/server/node_modules
 RUN rm -rf packages/frontend/node_modules
 RUN rm -rf packages/seanime-extensions/node_modules
 
-RUN pnpm install --prod --frozen-lockfile
+# The store cache mount matches the earlier install step, so production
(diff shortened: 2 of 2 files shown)


Qiiks:
Scope: split out of #1264 (Docker build caches). Includes the CodeRabbit fix from the #1264 review: the production `pnpm install --prod` step now mounts the same `/pnpm/store` cache as the dependency install. @coderabbitai review

Qiiks:
@coderabbitai full review

Qiiks:
Both findings fixed in 009d262:
- Dockerfile: added the `# syntax=docker/dockerfile:1` directive; every cache mount now has an explicit `id` (`pnpm-store`, `build-cache`) and `sharing=locked` so parallel builds on one builder can't corrupt a shared store.
- Cache poisoning (CWE-345): `.github/workflows/dev-docker.yml` now exports/imports `dev-`-prefixed gha cache scopes, so PR-triggered builds can never seed the deployment workflow's cache. (This was the right call — the dev workflow does accept arbitrary PR refs via workflow_dispatch.)

Qiiks:
@coderabbitai review

Qiiks:
On the CI cache-mount finding: correct in principle — `type=gha` layer caching does not persist cache-mount contents on ephemeral runners. Two reasons we left the workflows untouched here:

1. The cache mounts primarily target **self-hosted persistent builders** (e.g. a long-lived dockerd on a home server), where the mounts genuinely persist and dominate build times.
2. Wiring `buildkit-cache-dance` (or a persistent builder driver) into upstream's two workflows is a CI-infra change the maintainer may want to make differently — it felt out of scope for a Dockerfile PR, and we already had to touch `dev-docker.yml` for the security fix.

Happy to add the dance-based export/import to both workflows if you want CI speedups as part of this PR — say the word and we'll follow up.

Qiiks:
@coderabbitai review

PR #1305, opened 2026-09-10
feat: add optional yearless movie search fallback

ThePunpun:
Built-in movie title searches include the release year, which can miss releases whose indexed titles omit it.

Adds an optional fallback that retries movie title searches without the year when initial searches return too few unique results.

- Disabled by default, with a configurable result threshold (default: 3).
- Counts initial results before stream filtering, including IMDb results where the addon searches by both title and ID.
- Excludes results already rejected by addon-level checks, such as blacklisted categories, mismatched IMDb IDs, and unusable torrent entries.
- Evaluates addons independently, with per-indexer handling in Prowlarr.
- Preserves existing title limits, pagination, caching, and matching rules.
- Retains initial results if the fallback fails.
- Leaves ID-only searches and series searches unchanged.

Also fixes an existing Knaben deduplication bug: results were checked against the seen set but never added to it. This prevents duplicates from overlapping searches and applies even when fallback is disabled.

Validation: All 44 targeted tests, the core TypeScript build, and Prettier checks pass. In live scrape testing, final delivered streams increased from 1 to 3 after passing all downstream title-matching and stream filters.

## Summary by CodeRabbit

* **New Features**
  * Added optional yearless fallback searches for movies across built-in search providers when year-specific results are insufficient.
  * Added settings to enable the fallback and configure the minimum result threshold.
  * Added limits for Newznab/Torznab results and oversized responses.
  * Additional results are combined with initial results while avoiding duplicates.
  * Fallback errors are handled without losing original results.

* **Documentation**
  * Documented the new environment variables and default settings.

* **Tests**
  * Added comprehensive coverage for fallback behaviour and configuration.


ThePunpun:
@coderabbitai review

PR #1309, opened 2026-09-11
fix(streamnzb): read languages and parsedMediaInfo off StreamNZB streams

Gaisberg:
StreamNZB sends each stream's languages as ISO 639-1 codes in a `languages` field and a `parsedMediaInfo` block — the same shape the Newznab builtin fills from a feed's language and subs attributes. The StreamNZB preset used the generic parser, which derives languages from flag emojis alone and never fills subtitles, so every StreamNZB stream had "Unknown" subtitles and a required-subtitle filter dropped all of them.

Merge the codes into the parsed languages and take subtitles, languages and mediaInfoQuality from parsedMediaInfo when it is present.

## Summary by CodeRabbit

- **Improvements**
  - Stream information now includes normalised language codes.
  - Parsed files provide more complete media details, including language, subtitles, and quality.

### DIFF
--- a/packages/core/src/presets/streamnzb.ts
+++ b/packages/core/src/presets/streamnzb.ts
@@ -1,7 +1,21 @@
-import { Addon, Option, UserData, ParsedStream, Stream } from '../db/index.js';
+import {
+  Addon,
+  Option,
+  UserData,
+  ParsedStream,
+  Stream,
+  ParsedFile,
+} from '../db/index.js';
 import { Preset, baseOptions } from './preset.js';
 import { StreamParser } from '../parser/index.js';
-import { appConfig, createLogger, makeRequest } from '../utils/index.js';
+import {
+  appConfig,
+  createLogger,
+  makeRequest,
+  normaliseLanguage,
+  normaliseParsedMediaInfo,
+  ParsedMediaInfo,
+} from '../utils/index.js';
 import { constants } from '../utils/index.js';
 
 const logger = createLogger('streamnzb');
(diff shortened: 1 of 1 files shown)


packages/core/src/presets/streamnzb.ts
```diff
@@ -27,6 +41,55 @@ class StreamNZBStreamParser extends StreamParser {
...
+    const overrides: Partial<ParsedFile> = {};
+    if (provided.languages?.length) {
+      overrides.languages = [...provided.languages];
+    }
+    if (provided.subtitles?.length) {
+      overrides.subtitles = [...provided.subtitles];
+    }
+    if (provided.mediaInfoQuality) {
+      overrides.mediaInfoQuality = provided.mediaInfoQuality;
+    }
```
webstreamr, line 89: this looks weird to me, I think you want/need to set `overrides.mediaInfoQuality = 'addon'`  when either languages or subtitles are overridden. similar as e.g. meteor does it. alternatively, if we're sure that streamnzb is probing the streams, then it should be set to 'probe' instead.

packages/core/src/presets/streamnzb.ts
```diff
@@ -27,6 +41,55 @@ class StreamNZBStreamParser extends StreamParser {
...
+  // list of subtitle files — and without it every StreamNZB stream has
+  // "Unknown" subtitles and fails a required-subtitle filter.
+  protected override getParsedFileMergeOverrides(
+    stream: Stream,
+    _currentParsedStream: ParsedStream
+  ): Partial<ParsedFile> {
+    const provided = normaliseParsedMediaInfo(
+      (stream as Record<string, unknown>).parsedMediaInfo as
+        | ParsedMediaInfo
+        | undefined
```
webstreamr, line 77: I think it might be worth to re-visit all the types. if this was AI generated, maybe you can point it at other places and define a streamnzb return type or something similar also. in the end there should not really be a need for such unsafe type assertions

PR #1320, opened 2026-09-14
fix(builtins): preserve zero seeder counts

### LINKED ISSUE #1316: bug: seeder count of 0 is dropped and treated as missing in stream metadata
### Version

2026.09.09.1629-nightly (248d1c4a)

### Deployment

Self-hosted with Docker

### Area

Stream details (resolution, quality, language, release group parsed wrongly)

### Bug description

I found a difference between an explicitly reported seeder count of `0` and a genuinely missing seeder count.

I changed my custom formatter specifically to distinguish these two states:

- `Seeders: 0 reported` → when `stream.seeders` exists and its value is exactly `0`
- `Seeders: not reported` → when `stream.seeders` does not exist / was not provided

The formatter uses `stream.seeders::exists`, so it should be possible to distinguish a numeric zero from a missing value.

While testing this with **Red Notice (2021) / Alerta Vermelho (2021)**, I found that AIOStreams is losing an explicit `0` seeder value.

For the affected torrent, the upstream source (**CapybaraBR through Jackett/Torznab**) clearly reports:

```text
seeders = 0
```

I verified this directly in Jackett/CapybaraBR.

However, after the same torrent is processed by AIOStreams, `stream.seeders` behaves as if the value had never been provided at all.

Therefore my formatter displays:

```text
Seeders: not reported
```

instead of the expected:

```text
Seeders: 0 reported
```

Positive values do not appear to be affected. In the same CapybaraBR result list for **Red Notice (2021)**, other torrents report positive values such as `7`, `3`, and `2`, and those positive values are preserved correctly by AIOStreams in my testing.

This suggests that the problem is specifically related to numeric zero being treated as a missing/falsy value.

The custom formatter did not create the problem. It only made the difference between a missing value and an explicit numeric zero visible.

---

### Steps to reproduce

1. Run AIOStreams:

   ```text
   2026.09.09.1629-nightly (248d1c4a)
   ```

   Deployment:

   ```text
   Self-hosted with Docker
   ```

2. Configure the built-in **Torznab** integration using **Jackett** with the **CapybaraBR** indexer.

3. Use **TorBox** as the debrid service.

4. Search for:

   ```text
   Red Notice (2021)
   ```

   Portuguese title:

   ```text
   Alerta Vermelho (2021)
   ```

5. Find an **uncached** torrent returned by CapybaraBR that explicitly reports `0` seeders.

6. Check the same torrent directly in Jackett/CapybaraBR and confirm that the seeder column shows:

   ```text
   0
   ```

7. In AIOStreams, use a formatter that checks whether the `stream.seeders` field exists. For example:

   ```text
   {stream.seeders::exists["{stream.seeders}"||"not reported"]}
   ```

8. Search for **Red Notice (2021)** again through AIOStreams.

9. Observe that the torrent which Jackett/CapybaraBR explicitly reports with `0` seeders is displayed by AIOStreams as:

   ```text
   not reported
   ```

10. Compare it with torrents from the same CapybaraBR result set that report positive seeder counts.

    For example, other results in the same search report values such as:

    ```text
    7
    3
    2
    ```

    These positive values are preserved correctly by AIOStreams.

---

### Expected behaviour

An explicit numeric zero should remain a valid seeder value.

Expected data flow:

```text
Jackett / CapybaraBR
seeders = 0
        ↓
AIOStreams
stream.seeders = 0
        ↓
stream.seeders::exists = true
```

A genuinely missing seeder field should remain distinguishable:

```text
Jackett / upstream
seeders = missing
        ↓
AIOStreams
stream.seeders = undefined / missing
        ↓
stream.seeders::exists = false
```

In other words:

```text
seeders = 0
```

and:

```text
seeders = missing
```

should not represent the same state inside AIOStreams.

---

### Actual behaviour

Currently, an explicit upstream value of `0` becomes indistinguishable from a missing value:

```text
Jackett / CapybaraBR
seeders = 0
        ↓
AIOStreams
stream.seeders = missing / undefined
```

Therefore AIOStreams currently makes it impossible for a formatter or other stream-processing logic to distinguish between:

```text
0 seeders explicitly reported
```

and:

```text
seeder count not reported
```

This potentially affects more than only display formatting, because `stream.seeders` may also be used for:

- custom formatters
- seeder sorting
- seeder filters
- stream expressions
- debugging upstream indexer results

---

### Why I believe the value is lost inside AIOStreams

The same torrent can be checked directly in the upstream chain:

```text
CapybaraBR
    ↓
Jackett / Torznab
    ↓
AIOStreams
```

Jackett/CapybaraBR explicitly reports:

```text
seeders = 0
```

So this is not a case where the upstream source simply omitted the seeder count.

After AIOStreams processes the result, the zero value is no longer distinguishable from a missing value.

Because positive values such as `7`, `3`, and `2` are preserved correctly while `0` disappears, this looks consistent with a possible truthy/falsy check somewhere in the stream transformation/parsing path.

For example, logic equivalent to:

```ts
if (seeders) {
    // preserve the value
}
```

would preserve positive integers but discard numeric `0`, because `0` is falsy in JavaScript/TypeScript.

For a numeric field where zero is a valid value, the check should distinguish `0` from `undefined` or `null`.

---

### Additional verification

Image

Image

[aiostreams-config-GITHUB.json](https://github.com/user-attachments/files/32164826/aiostreams-config-GITHUB.json)

I attached screenshots showing the problem from both sides.

**Screenshot 1 — Jackett / CapybaraBR**

The affected **Red Notice (2021)** torrent explicitly shows:

```text
0 seeders
```

Other torrents in the same result list show positive values such as `7`, `3`, and `2`.

**Screenshot 2 — AIOStreams**

The same affected result is exposed by AIOStreams as having the seeder value:

```text
not reported
```

because `stream.seeders::exists` evaluates as though the value were absent.

The custom formatter is not generating or changing the seeder value. It only exposed that AIOStreams no longer has the explicit `0` value after processing the stream.

---

### Summary

Expected:

```text
0 reported upstream → stream.seeders = 0
missing upstream    → stream.seeders = undefined
```

Actual:

```text
0 reported upstream → stream.seeders = undefined / missing
missing upstream    → stream.seeders = undefined / missing
```

So an explicit seeder count of `0` is currently being collapsed into the same state as "seeder count not provided".

### Which addon or service is involved?

AIOStreams built-in addon: Torznab  Indexer: CapybaraBR through Jackett  The Torznab integration in AIOStreams is configured using: - Torznab URL provided by Jackett for the CapybaraBR indexer - API Key provided by Jackett  Debrid service: TorBox  The affected stream is returned through the following path:  CapybaraBR → Jackett → Torznab API → AIOStreams built-in Torznab addon → TorBox  I verified the affected result directly in Jackett, where the CapybaraBR result explicitly reports:  seeders = 0  However, after the result is processed by AIOStreams' built-in Torznab integration, the explicit numeric zero is no longer preserved and `stream.seeders` behaves as if the seeder count had not been provided.  Positive values from the same Jackett/CapybaraBR result set, such as 7, 3 and 2, are preserved correctly.  So the problem does not appear to be that Jackett or CapybaraBR failed to provide the seeder count. Jackett explicitly provides the value `0`; the distinction between `0` and a missing value is lost after the result enters AIOStreams.  TorBox is the configured debrid service, but this does not appear to be a TorBox issue because the incorrect/missing `stream.seeders` value is already observable in AIOStreams stream metadata.

### Exact filenames or stream data

```text
Affected title:
Red Notice (2021) / Alerta Vermelho (2021)

Affected release title as returned by Jackett/CapybaraBR:
2160p NF WEB-DL DDP5.1 Atmos HDR H.265-ABBIE DUAL-C.A.A

Upstream data observed directly in Jackett/CapybaraBR:
Seeders: 0

The Jackett/CapybaraBR result explicitly contains a numeric seeder count of 0.

After the same result is processed by AIOStreams, the seeder value is no longer available as `stream.seeders = 0`; it behaves as if the field were missing/undefined.

For comparison, other releases returned in the same Jackett/CapybaraBR search have positive seeder counts such as:

Seeders: 7
Seeders: 3
Seeders: 2

Those positive values are preserved by AIOStreams in my testing, while the explicit numeric value 0 is not.

Custom formatter check:
{stream.seeders::exists["{stream.seeders}"||"not reported"]}

For the affected release, this evaluates to:
not reported

Expected:
stream.seeders = 0

Actual:
stream.seeders = missing / undefined
```

### Configuration export

```json
Sanitized configuration export attached.
```

### Logs

```shell
Self-hosted with Docker.

No exception, HTTP error, playback error, or AIOStreams error stream is produced when reproducing this issue. The request completes successfully.

This is a stream metadata preservation issue rather than a request failure.

The relevant observed data flow is:

Jackett / CapybaraBR:
seeders = 0

AIOStreams after processing the same result:
stream.seeders = missing / undefined

Positive seeder values from the same Jackett/CapybaraBR search, such as 7, 3 and 2, are preserved normally.

The problem was reproduced while searching for:

Red Notice (2021) / Alerta Vermelho (2021)

Upstream:
Jackett → CapybaraBR → Torznab

AIOStreams integration:
Built-in Torznab addon

Debrid service:
TorBox

No proxy is involved in this reproduction.

There is no visible server-side exception associated with the affected stream; the issue is that the explicit numeric value `0` supplied by the upstream Torznab result is no longer present as `stream.seeders = 0` after AIOStreams processes it.

Server logs for the failing search request can be provided with LOG_LEVEL=debug if additional tracing is required.
```

### Device / client

Client used for reproduction: NuvioTV  Device: Philips 50PUG7408/78 OS: Android TV 14  This does not appear to be a client-side or playback issue.  The incorrect/missing `stream.seeders` value is already present in the AIOStreams stream metadata before the client displays the result. The client only renders the data returned by AIOStreams.  Therefore, the issue should be reproducible independently of the specific playback device/client.

### What I ruled out

- [x] The problem does NOT occur when using the upstream addon/service directly, without AIOStreams (or no upstream is involved).
- [ ] Playback issues — I tested on a second device/player and it fails there too.
- [ ] I retested with my stream expressions, regex filters, groups and catalog modifications disabled (where applicable) and the problem persists.
- [x] I retested with the proxy (MediaFlow/StreamProxy/built-in) disabled (where applicable) and the problem persists.

### Issue checklist

- [x] I have written a short, informative title that describes the actual problem.
- [x] I have searched open and closed issues and this is not a duplicate.
- [x] I have reproduced this on the [latest release](https://github.com/Viren070/AIOStreams/releases/latest) (or a current nightly).
- [x] I understand that issues missing the required information above may be closed without investigation.

procopio1000:
Fixes #1316.

When a built-in torrent result reports `seeders: 0`, `BaseDebridAddon._createStream()` drops the `👤 0` marker because its condition uses truthiness. `StreamParser` then returns `undefined`, so formatters and filters cannot distinguish a reported zero from missing metadata.

Check for a numeric seeder count instead. Zero and positive counts are preserved; missing or undefined counts remain omitted. Add regression tests that exercise the actual stream builder and parser, including a Usenet result without seeder metadata.

Validation:
- The zero-count regression fails on the unpatched version (`undefined !== 0`); the other four new cases pass.
- All 39 core tests pass with the fix, running transpiled test files against a clean, isolated Linux container built from `248d1c4` (with the compiled fixed module mounted).
- Core TypeScript build, type-check of the new test, Prettier, and `git diff --check` pass.

The Linux test run avoids the local Windows `yencode` native-build limitation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stream details now correctly display a seeder count of zero when applicable.
  * Seeder information is no longer shown when it is unavailable, preventing inaccurate values from being displayed.

* **Tests**
  * Added coverage for seeder handling across torrent and Usenet results, including zero, positive, and missing values.


packages/core/src/builtins/base/debrid.ts
```diff
@@ -907,7 +907,7 @@ export abstract class BaseDebridAddon<T extends BaseDebridConfig> {
     const name = `${torrentOrNzb.service?.library ? '🗃️ ' : ''}${isPrivate ? '🔑 ' : ''}[${shortCode} ${cacheIndicator}] ${this.name} ${isFreeleech ? 'FREELEECH' : ''} `;
     const description = `${torrentOrNzb.title ? torrentOrNzb.title : ''}\n${torrentOrNzb.file.name ? torrentOrNzb.file.name : ''}\n${
       torrentOrNzb.indexer ? `🔍 ${torrentOrNzb.indexer}` : ''
-    } ${'seeders' in torrentOrNzb && torrentOrNzb.seeders ? `👤 ${torrentOrNzb.seeders}` : ''} ${
+    } ${'seeders' in torrentOrNzb && typeof torrentOrNzb.seeders === 'number' ? `👤 ${torrentOrNzb.seeders}` : ''} ${
```
webstreamr, line 910: This makes total sense. Test might be a bit overkill, but the boss should be able to merge it I guess, no need to change :)

PR #1332, opened 2026-09-18
Fix anime NZB matching for bare absolute episode releases

ThePunpun:
Fixes anime Usenet releases that use a bare absolute episode number instead of standard season/episode notation.

Some NZB releases use formats such as:

[Group] Series Title 123 [CRC32]
GroupSubs.Series.Title.123.1080p

These releases could be rejected before normal stream matching because the generic parser does not always identify the title and bare episode number correctly.

This change:

Adds a conservative, metadata-anchored recovery path for anime NZB titles.
Requires an exact known metadata title and extracts the episode number from the release itself rather than copying the requested episode.
Treats recovered bare episode numbers as absolute/entry-relative episode numbers.
Validates recovered episodes against the requested absolute episode before accepting the NZB.
Applies the same recovery during Usenet inner-file selection so the correct file can be selected for playback.
Normalizes separator-only title differences such as spaces, dots, hyphens, and colons.
Supports adjacent release tags such as [720p][CRC32].
Prevents recovered episode mismatch state from leaking between streams that share the same stream ID.
Does not mutate cached parser results.
Remains scoped to anime Usenet/NZB processing; torrent/debrid matching is unchanged.

The recovery grammar is intentionally narrow and does not reinterpret releases with conflicting season/episode information, ranges, years, volumes, movie markers, or other ambiguous numeric values.

Tested with both absolute-episode metadata requests and season/episode metadata requests, including bare-number releases and standard season/episode releases.

## Summary by CodeRabbit

* **New Features**
  * Improved anime release recognition for NZB and Usenet streams, including titles with absolute episode numbers.
  * Anime metadata is now used to recover episode information when release titles are incomplete.
  * Recovered title, episode, and release-group details improve matching and playback selection.

* **Bug Fixes**
  * Incorrect absolute or relative episodes are now rejected more reliably.
  * Improved validation prevents unrelated or mismatched anime releases from being selected.


Spoorloos:
Been getting barely any usenet results for absolute numbered episodes (ahem One Piece) so I'd love to see this get merged!

PR #1335, opened 2026-09-18
Add SubRanker preset

iarhamanwaar:
Adds a preset for [SubRanker](https://github.com/iarhamanwaar/subranker) (MIT), a subtitle addon that sits in front of other subtitle addons rather than replacing them.

## What it does

It queries the subtitle addons you point it at, then ranks what comes back against the release actually being played, using release-name matching on both anime and scene/web naming. Along the way it prunes candidates that no longer download, and repairs the ones that are fixable: constant timing offsets and framerate drift, mojibake, embedded advertising, and cues that overlap so two speakers get drawn on top of each other.

It is complementary to the existing subtitle presets rather than competing with them, since they are what you would feed it.

## Notes on the implementation

**Self-hosted, so `URL: []` and the URL comes from the user.** I followed the pattern already used by `davex`, `usenetStreamer` and `streamnzb`. Rather than declaring a second `url` option, it overrides the one `baseOptions` already provides, marking it `required` and `showInSimpleMode`. `generateAddons` validates it and throws a readable error, matching how `usenetStreamer` handles a missing or malformed URL.

**Options map onto a config segment.** SubRanker carries per-user settings in a base64url segment on the path, so one instance serves everyone and there is nothing stateful to run. The option ids here correspond one-to-one with keys the addon already accepts.

Two behaviours worth flagging in review:

- If the user pastes a URL that already contains a config segment (what SubRanker's own setup page generates), it is passed through untouched, since it is already complete.
- If they paste a bare `…/manifest.json`, it is treated as the instance root and the options set here are still applied. I did this deliberately rather than returning it verbatim, because returning it verbatim silently discards every option the user just set.
- With no upstreams entered, it emits `{host}/manifest.json` and defers to whatever the instance was configured with, rather than sending an empty list that the instance would reject.

## Testing

Being straight about what I have and have not verified.

**Verified.** The preset is written against the actual `Option`, `Addon`, `PresetMetadata` and `baseOptions` signatures in this repo rather than copied from another preset by shape. I ran the `generateManifestUrl` logic against a live SubRanker instance's own route parser across every case a user could hit: plain host, trailing slash, pasted manifest URL, no upstreams, an upstream with a `/manifest.json` suffix, multiple upstreams, and an already-configured URL. All produce a URL the addon parses into the intended config.

**Not verified.** I have not been able to compile it inside the monorepo. Cloning this repo kept timing out on my connection, and I do not want to imply a typecheck I did not run. So please treat the type-level correctness as unconfirmed. If you have a build handy, that is the gap. I will fix whatever it reports, and I will confirm here once I have managed to build it locally.

Happy to adjust naming, option wording, category, or anything else that does not fit your conventions. And if you would rather not carry presets for addons without a public instance, no hard feelings — just say and I will close it.

https://claude.ai/code/session_0192bWBgbJmsnK7E58iWcggy

## Summary by CodeRabbit

- **New Features**
  - Added support for the self-hosted SubRanker subtitle-ranking add-on.
  - Added configuration options for the service URL, upstream sources, result limits, subtitle adjustments, OCR correction, and forced-subtitle handling.
  - Added validation for service addresses and automatic generation of configured add-on manifest URLs.
  - Registered SubRanker as an available preset.


iarhamanwaar:
Pushed a fix and closed most of the verification gap I flagged in the description.

**Fix.** `maxResults` had `constraints: { min: 0, max: 50 }`. Your `OptionDefinition` types `constraints.min` as `z.number().min(1)`, so `min: 0` would have failed validation. 0 is meaningful for this option — it means return everything — so the constraint is now upper-bound only, since your schema cannot express "0, or 1 to 50".

**Verification.** I still have not managed to clone the repo, so there is no compile. What I have now done instead is check every symbol and shape the preset depends on directly against `schemas.ts` and `preset.ts` on `main`:

- all four option `type` values used (`string`, `number`, `boolean`, `url`) are in the `OptionDefinition` enum
- `showInSimpleMode`, `required`, `default` and `constraints` all exist as I use them
- every field returned from `generateAddon` (`name`, `manifestUrl`, `enabled`, `library`, `resources`, `timeout`, `preset`, `headers`) is in `AddonSchema`, and `preset.id: ''` is accepted since that field has no `min`
- `PresetMetadataSchema` types `URL` as `z.array(z.string())`, which is why it is `[]` here
- `baseOptions` already supplies a `url` option, which is why this overrides it rather than declaring a second one with a duplicate id

That leaves genuine type inference as the only unchecked part. If anything there is wrong I will fix it promptly.

iarhamanwaar:
Closing the verification gap I flagged above. It now typechecks.

Since I still cannot clone the repo, I built a small harness instead: `Option`, `Addon`, `UserData`, `PresetMetadata`, `Resource`, `baseOptions` and the `Preset` base class transcribed by hand from `schemas.ts` and `preset.ts` on `main`, then compiled `subranker.ts` against them under `strict: true` with `moduleResolution: NodeNext`. Clean, including an assertion that the `METADATA` getter's return type is assignable to `PresetMetadata` (which is what the base class declares, so it is the constraint that actually matters).

I checked the harness has teeth rather than trusting a green exit code. Regressing `URL: []` back to `URL: ''` — the mistake I made in my first draft — produces exactly what it should:

```
error TS2322: Types of property 'URL' are incompatible.
  Type 'string' is not assignable to type 'string[]'.
error TS2417: Class static side 'typeof SubRankerPreset' incorrectly extends
  base class static side 'typeof Preset'.
```

**The honest caveat:** those are my transcriptions of your types, not your types. If I transcribed something too loosely, a real error could hide behind it. I was careful to carry the details that matter — `Option` as a loose object with the full `type` enum, `constraints` with optional `min`/`max`, `URL: string[]` on the metadata — but a build in your tree is still the authority. I would just rather say "typechecks against a faithful copy" than keep saying "unverified", and be clear about which one it is.

CodeRabbit's pass came back with no actionable findings.

